PR #23, opened 2019-11-20
Update 2019-11-21.md

glennblock:


IvanGoncharov:
@glennblock We had problem with Zoom here is new link: 

PR #26, opened 2019-11-20
Add Benjie to November WG

benjie:


IvanGoncharov:
@benjie We had problem with Zoom here is new link: 

PR #28, opened 2019-11-20
Add Jesse from Apollo to attendees for 2019-11-21.

abernix:


IvanGoncharov:
@abernix Merged, see you at the call 🎉

PR #32, opened 2019-11-21
Add myself to agenda

mike-marcacci:


IvanGoncharov:
@mike-marcacci We had problem with Zoom here is new link: 

PR #41, opened 2019-11-21
Add info about how the de-facto standard came to be

spawnia:
This prior art section seems useful as a starting point.


IvanGoncharov:
@spawnia Thanks a lot 👍

PR #43, opened 2019-11-22
Publish schedule and re-organize

sjparsons:
This PR publishes the general schedule for the working group and sets up a folder for future files related to the working-group itself.


IvanGoncharov:
@sjparsons Thanks a lot 👍

PR #60, opened 2020-01-27
Add myself to INTERESTED_DEVELOPERS.md.

abernix:
I've previously been interested and involved in previous meetings, but didn't notice this file until now.  I've therefore added myself in this commit! 😄 

Note: The instruction in the document states that we should order alphabetically by name though it doesn't include names, per se, just GitHub usernames.  I've therefore added myself alphabetized according to GitHub username, which seems to be the standard practice so far, but happy to adjust.

(Fwiw though, I won't be able to attend tomorrow's meeting. 😬)


benjie:
There's a meeting tomorrow?

PR #75, opened 2020-03-24
Tests to build

sjparsons:
The document included in this PR proposes a beginning set of tests based on the current draft spec that can be implemented. All comments are welcome as we start to evolve this list. Especially things that I have not yet included.

https://github.com/APIs-guru/graphql-over-http/issues/69

### DIFF
--- a/working-group/test-suite.md
+++ b/working-group/test-suite.md
@@ -1,12 +1,12 @@
 # Test Suite
 
-These are the description of the test that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation. 
+These are the descriptions of the test that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation and this document will be removed.
 
 ## Requests
 
-Server expense serialization format to be declared in the content-type of the request. 
+Server expects serialization format to be declared in the content-type of the request.
 
-Server should validate the content-type based in the request to ensure it's a content type (serialization format) that it supports 
+Server should validate the content-type based in the request to ensure it's a content type (serialization format) that it supports
 * Return `415 Unsupported Media Type` otherwise
 
 A GET method should be successful
--- a/working-group/test-suite.md
+++ b/working-group/test-suite.md
@@ -1,47 +1,58 @@
-# Test Suite
+# Automated Test Suite
 
-These are the descriptions of the test that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation and this document will be removed.
+> These are the descriptions of the tests that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation and this document will be removed.
+
+| Instruction in specification | Meaning | Result of test failure |
+|---|---|---|
+| MUST | Required behavior | ERROR
+| SHOULD | Recommended behavior | WARNING
+| MAY  | Allowed behavior  | NONE
+
+The tests described below indicate whether they will raise an ERROR or a WARNING if the assertion is violated. In order for a server to be marked as fully compliant it must receive no ERROR assertion failures when this automated test suite is run. In order to have a 100% scored-card the server must receive no ERROR or WARNING assertion failures when this automated test suite is run. 
 
 ## Requests
 
-Server expects serialization format to be declared in the content-type of the request.
+A GET method SHOULD be successful
+* Send a GET HTTP request for introspection
+  * **WARNING** Assert that the server returns a 200 HTTP response
 
-Server should validate the content-type based in the request to ensure it's a content type (serialization format) that it supports
-* Return `415 Unsupported Media Type` otherwise
 
-A GET method should be successful
-* Attempt a GET method request for introspection
-   query (serialized as URL encoded key-value pair)
+A GET method with `query`, `operationName` and `variables` SHOULD be successful
+* Send a GET HTTP request for introspection using `query`, `operationName`, and `variables`. These should encoded as key-value pairs as query params. `variables` must be a URL-encoded valid JSON string.
+  * **WARNING** Assert that the server returns a 200 HTTP response
 
-A POST method should be successful
-* Attempt a POST method for introspection
-   query (serialized in body)
+At least one of either a POST or a GET method MUST be successful
+* Send a GET HTTP request for introspection AND send a POST HTTP request for introspection:
+  * **ERROR** Assert that the server returns a 200 HTTP response to one of the requests
 
-A GET method with query, operationName and variable should be successful
+A POST method SHOULD be successful
+* Send a POST HTTP request for introspection
+  * **WARNING** Assert that the server returns a 200 HTTP response
 
-A POST method with query, operationName and variable should be successful
+A POST method with `query`, `operationName` and `variables` SHOULD be successful
+* Send a POST HTTP request for introspection using `query`, `operationName`, and `variables`. These will be encoded as a JSON object.
+  * **WARNING** Assert that the server returns a 200 HTTP response
+
 
 ## Responses
 
-A 200 response should include a body. The body should include at least a `data` attribute or an `errors` attribute.
+A 200 HTTP response to a request without a specified content-type MUST include a body and MUST indicate the encoding with the `content-type` HTTP header.
+* Send a GraphQL query in an HTTP request without an `Accept` header. May be a GET or POST, based on which is supported by this server.
+  * **ERROR** Assert that the HTTP response includes an HTTP Header `Content-type: application/graphql+json`
+  * **ERROR** Assert that the HTTP response body is valid JSON containing at least a non-null `data` or a non-null `error` top-level attribute.
+    * This header MAY include encoding information (e.g. `Content-type: application/graphql-json; charset=utf-8`)
 
-If no `Accept` header is sent in the request the server MAY choose whatever serialization format it defaults to. 
+Servers MUST respect the content type requested in the `Accept` HTTP header in the request.
+* Send a GraphQL query in an HTTP request with HTTP Header `Accept: application/graphql+json`.
+  * **ERROR** Assert that the HTTP response includes an HTTP Header `Content-type: application/graphql+json`
+  * **ERROR** Assert that the HTTP response body is valid JSON
+* Send a GraphQL query in an HTTP request with HTTP Header `Accept: unknown/content-type`, and invalid and unsupported content-type
+  * **ERROR** Assert that either:
+    * The HTTP response is `406 Not Acceptable`; OR
+    * The HTTP response includes an HTTP Header `Content-type: application/graphql+json` and the HTTP response body is valid JSON
 
-The server SHOULD include a `content-type` header in the response to indicate what serialization format was used in the response.
-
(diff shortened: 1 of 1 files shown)


working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
+# Test Suite
+
+These are the description of the test that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation. 
+
+## Requests
+
+Server expense serialization format to be declared in the content-type of the request.
```
sungam3r: Expense?

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+* Attempt a POST method for introspection
+   query (serialized in body)
+
+A GET method with query, operationName and variable should be successful
+
+A POST method with query, operationName and variable should be successful
+
+## Responses
+
+A 200 response should include a body. The body should include at least a `data` attribute or an `errors` attribute.
```
sungam3r: Are we talking about json responses? It should be stated explicitly. If so then I propose to change wording from attribute to property.

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+   the response have a media type that is listed as acceptable, the
+   origin server can either honor the header field by sending a 406 (Not
+   Acceptable) response or disregard the header field by treating the
+   response as if it is not subject to content negotiation.
+   [HTTP 1.1 Accept](https://tools.ietf.org/html/rfc7231#section-5.3.2)
+
+
+Server may response with non-200 status codes
+
+Syntax error in the query should result in either a 200 with an `error` attribute in the body or a 400 response.
```
sungam3r: For syntax error spec now allows only 4xx and 5xx error codes
https://github.com/APIs-guru/graphql-over-http#status-codes

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
+# Test Suite
+
+These are the description of the test that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation.
```
benjie:
```suggestion
These are the descriptions of the tests that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation. 
```

working-group/test-suite.md
benjie:
```suggestion
Server expects serialization format to be declared in the `Content-Type` header of the request.
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
+# Test Suite
+
+These are the description of the test that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation. 
+
+## Requests
+
+Server expense serialization format to be declared in the content-type of the request. 
+
+Server should validate the content-type based in the request to ensure it's a content type (serialization format) that it supports
```
benjie:
```suggestion
Server SHOULD validate the request's `Content-Type` to ensure it is supported
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
+# Test Suite
+
+These are the description of the test that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation. 
+
+## Requests
+
+Server expense serialization format to be declared in the content-type of the request. 
+
+Server should validate the content-type based in the request to ensure it's a content type (serialization format) that it supports 
+* Return `415 Unsupported Media Type` otherwise
```
benjie:
```suggestion
* Return `415 Unsupported Media Type` if request `Content-Type` is not supported
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+
+A GET method should be successful
```
benjie:
```suggestion
A GET method SHOULD be successful
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+* Attempt a GET method request for introspection
+   query (serialized as URL encoded key-value pair)
+
+A POST method should be successful
```
benjie:
```suggestion
A POST method SHOULD be successful
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+* Attempt a POST method for introspection
+   query (serialized in body)
+
+A GET method with query, operationName and variable should be successful
+
+A POST method with query, operationName and variable should be successful
```
benjie:
```suggestion
At least one of the following MUST be successful:
* A GET method with query, operationName and variables
* A POST method with query, operationName and variables
```

working-group/test-suite.md
benjie:
```suggestion
A 200 response MUST include a body. If the body has `Content-Type` of `application/json`, `application/graphql` or `application/graphql+json`, the body MUST be a JSON object that conforms to the ["Response Format" as specified in the GraphQL Specification](https://spec.graphql.org/June2018/#sec-Response-Format).
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Responses
+
+A 200 response should include a body. The body should include at least a `data` attribute or an `errors` attribute.
+
+If no `Accept` header is sent in the request the server MAY choose whatever serialization format it defaults to. 
+
+The server SHOULD include a `content-type` header in the response to indicate what serialization format was used in the response.
```
benjie:
```suggestion
The server MUST specify a valid `Content-Type` header in the response, indicating which serialziation format was used. This header MAY include encoding information (e.g. `Content-type: application/json; charset=utf-8`)
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+
+If `Accept` header is given in the request, the server should serialize the body in the response according to one of the content-types in the `Accepted` HTTP header.
+
+If all of the content-types listed in `Accept` are not supported by the server there are two valid options: return a 406 or return in the default content-type.
```
benjie:
```suggestion
The server SHOULD validate the `Accept` header, if any, in the request.

The server SHOULD serialize the body in the response according to the first supported content-type in the `Accept` header, if present.

If none of the content-types in the `Accept` header are supported, the server MUST _either_:
* return a `406 Not Acceptable` status code; _or_
* ignore the `Accept` header and instead use the server's preferred `Content-Type`.
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+> If the header field is
+   present in a request and none of the available representations for
+   the response have a media type that is listed as acceptable, the
+   origin server can either honor the header field by sending a 406 (Not
+   Acceptable) response or disregard the header field by treating the
+   response as if it is not subject to content negotiation.
+   [HTTP 1.1 Accept](https://tools.ietf.org/html/rfc7231#section-5.3.2)
+
+
+Server may response with non-200 status codes
```
benjie:
```suggestion
Server MUST NOT respond with any status code between 201 and 299 inclusive.

Server MUST respond with a 200 status code if the GraphQL request has been executed [as defined in "Executing Operations" in the GraphQL Specification](https://spec.graphql.org/June2018/#sec-Executing-Operations).

Server MAY respond with non-200 status codes if the GraphQL request is invalid, or if the server is unable to process it.
```

working-group/test-suite.md
spawnia: I think `application/graphql` is not meant to be used for responses, only for requests. It makes sense when sending a raw query string like so:

    curl -v -H "Content-Type: application/graphql" -d "{ hello }"  "localhost:3000/graphql"

It might be a valid response header, but not to represent the result of a GraphQL operation. I guess it could be used for something like a schema/query generator.

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
+# Test Suite
+
+These are the descriptions of the test that we intend to build. This document is a scaffold until we have the tests implemented. At that point, the code itself will become the documentation and this document will be removed.
+
+## Requests
+
+Server expects serialization format to be declared in the content-type of the request.
+
+Server should validate the content-type based in the request to ensure it's a content type (serialization format) that it supports
```
spawnia, line 9:
```suggestion
Server SHOULD validate the content-type based in the request to ensure it's a content type (serialization format) that it supports
```

working-group/test-suite.md
```diff
@@ -0,0 +1,47 @@
...
+
+A GET method should be successful
+* Attempt a GET method request for introspection
+   query (serialized as URL encoded key-value pair)
+
+A POST method should be successful
+* Attempt a POST method for introspection
+   query (serialized in body)
+
+A GET method with query, operationName and variable should be successful
```
spawnia, line 20:
```suggestion
A GET method with query, operationName and variable SHOULD be successful
```

working-group/test-suite.md
sjparsons: I believe the content-type `application/graphql+json` can refer to a GraphQL JSON response document and the GraphQL request body. In other words, I think it should be a JSON document containing at least one of `data` or `errors` as top level attributes in a response. 

When sending a POST request, a client should send `application/graphql+json` in the general case because they are also sending a JSON document, this time including a `query`, `operationName` and `variables` as the top-level attributes. The GraphQL query itself, which is found in this object (in the case of a POST request) at the `query` parameter should conform to the GraphQL specification.

I'm not clear on whether `application/graphql` is a synonym for `application/graphql+json`. Hypothetically, `application/graphql` could be distinct and refer to a valid GraphQL document string. However, we then would not be able to use it in GraphQL over HTTP since at no time do we send _only a GraphQL query string_ to a server or receive _only a GraphQL query string_ from a server.

working-group/test-suite.md
sjparsons: We need to put this language in the spec first and then let the test cases fall out from that. Could this be made as a proposal on https://github.com/APIs-guru/graphql-over-http/pull/79?

working-group/test-suite.md
sjparsons: These are some good suggestions, but I feel they pertain more to the spec than to the tests we wish to write. I've attempted to work them into this PR: https://github.com/APIs-guru/graphql-over-http/pull/83

working-group/test-suite.md
sjparsons: That's a good point. I feel conflicted over this since I think the spec should probably be more permissive on this point. But my concern is better to raise on the main spec, than here.

sjparsons:
I've made some substantial changes to take into account most of the feedback. Some of this has resulted in another PR since I think some of the great suggestions here are spec-worthy, not just relevant for our tests. 

I've also increased the clarity of the difference between ERROR and WARNING assertion failures. Hopefully those can be worked into our tests.

working-group/test-suite.md
sjparsons: I added some of this in #83

working-group/test-suite.md
sungam3r: The point is that 200 may be returned only when GraphQL execution actually started. Otherwise there is no "GraphQL context". Syntax error or other errors **before** execution should return 4xx or 5xx.

PR #77, opened 2020-03-24
Fix time of meetings to align with our plan

sjparsons:
This month I made a mistake and set up the meeting for 1700 instead of 1900 UTC. One of our members, @jaydenseric in Australia wasn't able to join since it was 4am. I'm correcting this now so that we get it right in the future.


jaydenseric:
Thanks Sam, I'm really sad I missed it! I woke up early and was at my desk and ready for the usual 6am Melbourne Australia time, then realised the meeting had already happened 2 hours earlier 😳

sjparsons:
So sorry @jaydenseric!

spawnia:
What a pity, would have loved to have you in. Looking forward to next time.

PR #79, opened 2020-04-19
Add clarifications regarding HTTP status codes

spawnia:
As discussed in the 2020-03-24 working group meeting, i expanded the section around status codes with clarifications and explanations.

See https://github.com/APIs-guru/graphql-over-http/issues/76


README.md
```diff
@@ -159,15 +159,22 @@ any errors encountered during the request.
...
-If the `data` entry in the response has any value other than `null` (when the operation has successfully executed
-without error) then the response should use the 200 (OK) status code.
-If the operation failed before or during execution, due to a syntax error, missing information, validation error
-or any other reason, then response should use the 4XX or 5XX status codes.
+A server must differentiate between different error conditions.
+
+The result of executing a GraphQL operation may contain partial data as well as encountered errors.
+Errors that happen during execution of the GraphQL operation become part of the result.
+If the server successfully executed parts of the operation and returns a well-formed response
+that contains the key `data`, the status code MUST be `200` (OK).
```
spawnia, line 174: I am not sure yet on how we should handle the case where `data == null`. Here is a relevant section of the GraphQL spec:

> If an error was encountered before execution begins, the data entry should not be present in the result.
>
> If an error was encountered during the execution that prevented a valid response, the data entry in the response should be null.

What would be examples of errors during execution that prevent a valid response?

sungam3r: Response serialization errors.

benjie: If there were an non-null root field that threw an error.

spawnia: Do you think it would be appropriate to return a non-200 status code for those cases?

benjie: For my one, no. But the reference implementation does, I think? @IvanGoncharov  is on the fence about it, I think? (Sorry for all the "I think"s!)

benjie: Response serialization errors, e.g. happens after GraphQL resolves but for some reason cannot write the result out, should be a 500 to my mind.

spawnia: @benjie i agree, errors on non-null field are still within the realm of GraphQL execution. The behaviour is consistent with what happens within nested fields. The response is still well-formed GraphQL, including appropriate errors.

Returning a different status code would be the kind of special casing that we generally want to avoid.

README.md
```diff
@@ -159,15 +159,25 @@ any errors encountered during the request.
...
-If the `data` entry in the response has any value other than `null` (when the operation has successfully executed
-without error) then the response should use the 200 (OK) status code.
-If the operation failed before or during execution, due to a syntax error, missing information, validation error
-or any other reason, then response should use the 4XX or 5XX status codes.
-It is recommended to use the same error codes as the [reference implementation](https://github.com/graphql/express-graphql).
+The result of executing a GraphQL operation may contain partial data as well as encountered errors.
+Errors that happen during execution of the GraphQL operation become part of the result.
+
+If the server successfully executed parts of the operation and returns a well-formed response
+that contains a non-null key `data`, the status code SHOULD be `2xx`.
```
spawnia, line 173: Still on the fence about this part, i think it depends upon what we can do with the response Content-Type.

When we have an unambigious Content-Type such as `application/graphql+json`, we can safely use any `2xx` or even other status codes for valid responses. That would be my preferred option.

If not, we might have to say the status code MUST be exactly `200` for valid responses.

benjie: This should be `MUST` IMO.

ralfhandl: Looking at https://tools.ietf.org/html/rfc7231#section-6.3 I also think it MUST be `200`. 

- `201 Created` would seem only appropriate if the operation created a resource either at the request URL or at a location returned in a `Location` header - that seems weird in a GraphQL context.''

- `202 Accepted` would be appropriate if the response is a partial result, sent while the server is working on remaining parts of the operation. 

  That would actually be an interesting feature, but how is then the next "chunk" of results returned?

- `203 Non-Authoritative Information` seems rather special, and I haven't seen that used so far.

- `204 No Content` disqualifies itself 😄 

- `205 Reset Content` also has no content in the response.

ghmcadams: `201 Created ` might make sense for some mutations.  Otherwise, `200` is the only status code that seems appropriate for a query.

README.md
```diff
@@ -159,15 +159,25 @@ any errors encountered during the request.
...
+
+If the status code is `200`, the response MUST be a well-formed GraphQL response.
+
+The server SHOULD respond with status code `400` (Bad Request) in case of a validation error that
+prevented execution of the request.
+
+The server SHOULD respond with the appropriate status code depending on the concrete error condition,
+for example `401` (Unauthorized) or `500` (Internal Server Error).
```
ghmcadams, line 181: These status codes should only apply if they are absolute (which most likely means the error occurred before the processing of the GraphQL request).  If an error occurs (either `401` or `500`) inside of a resolver or a directive's function, then a `200` should be sent, with an error key present.

README.md
spawnia: I guess the Content-Type does not necessarily say that the request was successful.

We might want to reserve additional status codes such as `202 Accepted` for features such as deferred queries, where the result might be valid GraphQL - but with some extra bits tacked on.

If a great use case for different status codes comes up, e.g. `201` for some mutations proves to be valuable information, we can open up the spec at a later point to allow it.

As of now, i think GraphQL does quite well by including extra information as part of `data`, `errors` or `extensions`, without leaning on status codes. The simplicity and interoperability with existing tools is another great reason for going `200` only for actual GraphQL responses.

README.md
spawnia: Agreed, i clarified that those status codes should only be returned for errors that completely prevent the successful execution of the request.

README.md
ghmcadams: I totally agree.  I like the idea of reserving other codes for future potential use.  We will definitely have to think about how to convey a partial response.

README.md
```diff
@@ -161,23 +161,24 @@ any errors encountered during the request.
...
-prevented execution of the request.
+The result of executing a GraphQL operation may contain partial data as well as encountered errors.
+Errors that happen during execution of the GraphQL operation become part of the result.
+
+In case of errors that completely prevent the successful execution of the request, the server SHOULD respond
+with the appropriate status code depending on the concrete error condition. For example:
 
-The server SHOULD respond with the appropriate status code depending on the concrete error condition,
-for example `401` (Unauthorized) or `500` (Internal Server Error).
+- `400` (Bad Request), in case of a validation error
```
ghmcadams, line 180: I believe that in our last group conversation, we did not reach consensus on whether validation errors should respond with a 400.

README.md
```diff
@@ -161,23 +161,24 @@ any errors encountered during the request.
...
+- `401` (Unauthorized), if the client is not allowed to access the schema or any of the requested fields
```
ghmcadams, line 181:
```suggestion
- `401` (Unauthorized), if an unauthenticated client is not allowed to access the schema
```

README.md
```diff
@@ -159,15 +159,26 @@ any errors encountered during the request.
...
+The result of executing a GraphQL operation may contain partial data as well as encountered errors.
+Errors that happen during execution of the GraphQL operation become part of the result.
+
+In case of errors that completely prevent the successful execution of the request, the server SHOULD respond
+with the appropriate status code depending on the concrete error condition. For example:
+
+- `400` (Bad Request), in case of a validation error
+- `401` (Unauthorized), if the client is not allowed to access the schema or any of the requested fields
+- `500` (Internal Server Error), for unexpected failures in the server
+
```
ghmcadams, line 183: Do we want to also mention `403 (Forbidden)`?  This may be contentious in that authorization could be done at the field level.  I argue that some servers may also apply it at the schema/server level.

spawnia: Right, i don't think the examples need to be exhaustive. As you noted, this one is a bit tricky and might require a somewhat lengthy explanation in order to not be misleading.

On the other hand, authorization might be a great and illustrative example of an error condition that can be handled on the field level or on the HTTP level.

spawnia: https://github.com/APIs-guru/graphql-over-http/pull/79/commits/8a1108f0f5395e9f257dbaab050a0e4c18c08e94

README.md
```diff
@@ -159,15 +159,34 @@ any errors encountered during the request.
...
-or any other reason, then response should use the 4XX or 5XX status codes.
-It is recommended to use the same error codes as the [reference implementation](https://github.com/graphql/express-graphql).
+If the server successfully executed parts of the operation and returns a well-formed response
+that contains a non-null key `data`, the status code MUST be `200`.
+
+If the status code is `200`, the response MUST be a well-formed GraphQL response.
+
+The result of executing a GraphQL operation may contain partial data as well as encountered errors.
+Errors that happen during execution of the GraphQL operation typically become part of the result,
+as long as the server is still able to produce well-formed response.
```
ralfhandl, line 176:
```suggestion
as long as the server is still able to produce a well-formed response.
```

README.md
spawnia: This example section is prefaced with

> In case of errors that completely prevent the successful execution of the request

There might be validation errors that only cause a partial error, those would not fall into this category of errors.

benjie: What validation errors would only cause a partial error? Is that `execute`-time validation of variables, perhaps?

spawnia:
```graphql
mutation turnUpTheHeat(degrees: 100)
```

```json
{
  "data": {
    "turnUpTheHeat": 80
  },
  "errors": [{ "message": "That's too hot", ... }]
}
```

README.md
```diff
@@ -159,15 +159,34 @@ any errors encountered during the request.
...
+- `400` (Bad Request), in case of a validation error
+- `401` (Unauthorized), if an unauthenticated client is not allowed to access the schema
+- `500` (Internal Server Error), for unexpected failures in the server
+
+As an example, the way authorization errors may be handled illustrates the different error conditions.
+If a client is forbidden from making any GraphQL requests at all, the server may respond with
+status code `403` (Forbidden) and not return any GraphQL data.
+In contrast to that, a client might be allowed to access the schema, but requests some restricted fields.
+Given the server can provide at least partial `data`, the response would have status code `200` (OK) and 
+include `errors` as part of the GraphQL response.
```
ghmcadams, line 190: This is awesome!

README.md
benjie: The way I see it, we have three different types of validation errors:

1. GraphQL Specification document validation errors (valid syntax, no loops in fragments, etc)
2. Custom document validations that can be performed before execution (e.g. [graphql-depth-limit](https://github.com/stems/graphql-depth-limit/))
3. Runtime validations performed by resolvers (i.e. during the execution of the GraphQL operation).

I group 1 and 2 together, and those should result in a 400.

Type 3 should NOT result in a 400, these are errors that occur during execution, so it's a partial success. Other fields in the query could execute independently and successfully, so using a 400 status goes against the partial-success ideal of GraphQL, and determining if there _were_ any additional fields seems beyond the scope of this spec.

(Aside: I edited the word "query" out of this comment in so many places! :sweat_smile:)

benjie: Maybe we can summarise this by changing "validation" to "GraphQL document validation"?

```suggestion
- `400` (Bad Request), in case of a GraphQL document validation error
```

spawnia: There is another type (maybe type 0?), for instance malformed JSON or typos like `{"qurey":...`. Those don't have anything to do with GraphQL documents.

The discussion we are having shows there is plenty of room for ambiguity and leaves implementors guessing. We could at least add a non-normative example of how different validation errors might be handled, akin to the example for 403.

benjie: You're totally right. I've added type 0 to my comment above.

README.md
```diff
@@ -159,15 +159,34 @@ any errors encountered during the request.
...
+If the server successfully executed parts of the operation and returns a well-formed response
+that contains a non-null key `data`, the status code MUST be `200`.
+
+If the status code is `200`, the response MUST be a well-formed GraphQL response.
+
+The result of executing a GraphQL operation may contain partial data as well as encountered errors.
+Errors that happen during execution of the GraphQL operation typically become part of the result,
+as long as the server is still able to produce a well-formed response.
+
+In case of errors that completely prevent the successful execution of the request, the server SHOULD respond
```
jaydenseric, line 178: As discussed in today's working group meeting, it is my belief this should be a `MUST`.
```suggestion
In case of errors that completely prevent the successful execution of the request, the server MUST respond
```

We should change the following phrasing from using the word "example", to concretely outlining the situations the status is `400`, or `200` — as outlined by @benjie here: https://github.com/APIs-guru/graphql-over-http/pull/79/files#r423698052 .

README.md
```diff
@@ -175,18 +175,26 @@ The result of executing a GraphQL operation may contain partial data as well as
...
+status code `400` (Bad Request). Types 3. and 4. may still allow partial execution and
+MUST result in status code `200`.
+
+If an unauthenticated client is not allowed to access the schema, the server MUST respond with
+status code `401` (Unauthorized).
+
+If the server failed unexpectedly, it MUST respond with status code `500` (Internal Server Error).
+
+If a client is not authorized to access any of the requested fields, the server MUST respond with
+status code `403` (Forbidden) and not return any `data`.
```
benjie, line 198: Requiring that we check that each individual field requested was forbidden is quite a burden, and those constraints live in the GraphQL schema itself rather than the GraphQL server, so enforcing them will be challenging (e.g. determining the difference between a field failing due to it being forbidden, versus it failing just due to an error occurring). Instead, 403 should be used when access to the GraphQL schema itself is denied.

I've also added an extra paragraph to deal with the two `SHOULD`s.

```suggestion
If a client is not authorized to access the GraphQL schema, the server SHOULD respond with
status code `403` (Forbidden) and not return any `data`.

If a client is not permitted to access the GraphQL schema for any reason, the server MUST respond with a `4xx` or `5xx` status code (a `2xx` status code MUST NOT be used).
```

README.md
```diff
@@ -175,18 +175,26 @@ The result of executing a GraphQL operation may contain partial data as well as
...
+3. Custom document validations that can be performed before execution (e.g. graphql-depth-limit)
+4. Runtime validations performed by resolvers (i.e. during the execution of the GraphQL operation).
 
-It is RECOMMENDED to use the same error codes as the [reference implementation](https://github.com/graphql/express-graphql).
+The error types 1. and 2. prevent execution of the request entirely and MUST result in
+status code `400` (Bad Request). Types 3. and 4. may still allow partial execution and
+MUST result in status code `200`.
+
+If an unauthenticated client is not allowed to access the schema, the server MUST respond with
+status code `401` (Unauthorized).
```
benjie, line 193: At the moment this precludes the use of 403 for unauthenticated clients. Further, one of 401, 403 or 404 should be permitted in these situations, they each have different security implications.

```suggestion
If an unauthenticated client is not allowed to access the schema, but an authenticated client might be permitted, the server SHOULD respond with
status code `401` (Unauthorized).
```

README.md
spawnia: Included the error types within the spec text.

README.md
spawnia: Suppose a server implements field level authorization, and the client accesses some allowed fields and some disallowed fields. The responses might contain partial `data`, as well as a bunch of `errors`. That MUST definitely be a 200.

However, when all requested fields are forbidden, the server is unable to produce any `data`. Consider the above section:

> In case of errors that completely prevent the successful execution of the request, the server MUST respond
with the appropriate status code depending on the concrete error condition.

The requests was not successful, the client did not get any of the data they asked for. I think that should be reflected in a non-200 status code.

> enforcing them will be challenging

Agreed, no one said this was going to be easy ;) In all seriousness, it seems doable. A well-built GraphQL server should be able to make it happen. If not, well, then it is fine if they are non-compliant with this aspect of the spec.

README.md
spawnia: > At the moment this precludes the use of 403 for unauthenticated clients.

Missing or failed authentication is precisely what 401 is meant for, or is it not? 403 and 404 each mean something different.

Concerning the use of `SHOULD` or `MUST`: How about we use `MUST` for things that are important for interoperability with clients? In this case, it seems like clients might be able to do something useful with a 401, e.g. redirecting to a login page.

README.md
benjie: I think that the spec should be written in a way that makes compliance easy - rather than talking about it in terms of "allowed fields and disallowed fields" which is not a concept that exists in the GraphQL spec, we should talk about it in terms of successful execution. In this case what you're really saying is "should resolving the request with graphql result in the data field being null, a 4xx or 5xx status should be used".

It's my belief that you should be able to build a GraphQL server without needing to know about the internal logic of the underlying GraphQL schema - everything the server needs to know should be available as a result of calling the `parse`, `validate` and `execute` GraphQL functions in whatever language you're using; these functions do not have any concept of "field-level authorization".

README.md
benjie: I should expand on the security implications I mentioned. Statuses 401 and 403 (generally) confirm the resource exists and that you're not allowed to see it, 404 does not. So some servers choose to use 404 to get bots/scripts to "stop looking" - the page doesn't seem to exist so there's no point trying to hack it. Have a read of: https://www.bennadel.com/blog/2400-handling-forbidden-restful-requests-401-vs-403-vs-404.htm

It should be up to the developer to decide which of these status codes is relevant to their API design.

README.md
ghmcadams: I agree.  The server (the part that is outside of the GraphQL implementation) should not know or care about any part of the schema or any rules associated with it.  Though the server is permitted to determine whether a user is authorized to access the schema, and when not MUST respond with a 403 status code.  Equally, if the server decides that a user must be authenticated in order to access the schema, and an unauthenticated user makes a request, it MUST respond with a 401 status code.

If the GraphQL implementation successfully executed a request (even if all requested data is protected and the requesting user has no access to any of it), the server MUST respond with a 200 status code.

benjie:
Agreed in working group:

**If the server returns GraphQL Content-Type and contains a non-null "data" entry, then it MUST use 2xx status code and it SHOULD use 200 status code.**

benjie:
Generally agreed in working group:

**If the server returns GraphQL Content-Type and status code is not 2xx, "data" entry must be either null or not present (depending on if it were an execution error or a validation error)**

sjparsons:
_For documentation_

Action from 2020-06-30 meeting: 
> Split the PR, keeping the minimal agreed upon changes in the original and open new ones for more controversial changes.

spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,17 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
-If the operation failed before or during execution, due to a syntax error, missing information, validation error
-or any other reason, then response should use the 4XX or 5XX status codes.
-It is recommended to use the same error codes as the [reference implementation](https://github.com/graphql/express-graphql).
+If the response has Content-Type GraphQL and contains a non-null `data` entry,
+then it MUST have status code `2xx`, and it SHOULD have status code `200`.
+
+If the response has Content-Type GraphQL and has a non-`2xx` status code,
+the `data` entry must be either:
+- equal to `null` (in case of an execution error)
+- not present (in case of a validation error)
```
IvanGoncharov, line 210: The value of `data` is fully covered by GraphQL specification I don't think we should duplicate it here.
Instead, just specify it should be `null` or not present at all.

spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,17 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
+
+The result of executing a GraphQL operation may contain partial data as well as encountered errors.
+Errors that happen during execution of the GraphQL operation typically become part of the result,
+as long as the server is still able to produce a well-formed response.
```
sjparsons, line 214: @IvanGoncharov commented in the working group that this paragraph would be better as a non-normative note since this is not a requirement.

sjparsons:
@spawnia we've agreed to merge this PR once the two comments above are resolved. Thanks for your work on this!

spawnia:
@sjparsons @IvanGoncharov the comments are resolved, this is good to be merged now.

spawnia:
@mmatsa

PR #94, opened 2020-04-30
Clarify URL requirements

spawnia:
Relates to the discussion in https://github.com/APIs-guru/graphql-over-http/issues/91

@ghmcadams  @ralfhandl

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,13 +78,15 @@ providing a GraphQL request in an HTTP request and GraphQL response in an HTTP r
 
 A GraphQL over HTTP compliant server MUST designate at least one URL that handles GraphQL requests.
 
-It is allowed for the same GraphQL schema to be available on multiple URLs.
-
-All GraphQL operations that are available in the schema MUST be available on each designated URL.
+All GraphQL operations that are available in a schema MUST be available on all URLs that are designated to it.
 That means the client can send all their GraphQL requests to a single endpoint, regardless
 of the contained Operations, Fragments or Variables.
 
-Other protocols may also use those URLs.
+Multiple URLs MAY exist on a server to handle GraphQL requests, potentially serving different
+but independent schemas. It is allowed for the same GraphQL schema to be available on multiple URLs.
+
+Other protocols or services may also use those URLs, as long as they don't conflict with the
+server's responsibility to handle GraphQL requests.
 
 It is RECOMMENDED to end the path component of the URL with `/graphql`, for example:
 - `http://example.com/graphql`
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ of the contained Operations, Fragments or Variables.
 Multiple URLs MAY exist on a server to handle GraphQL requests, potentially serving different
 but independent schemas. The same GraphQL schema MAY be available on multiple URLs on the server.
 
-Other protocols or services may also use those URLs, as long as they don't conflict with the
+Those URLs MAY also be used for other purposes, as long as they don't conflict with the
 server's responsibility to handle GraphQL requests.
 
 It is RECOMMENDED to end the path component of the URL with `/graphql`, for example:
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ A conforming implementation of GraphQL over HTTP may provide additional function
 but must not where explicitly disallowed or it would otherwise result
 in non-conformance.
 
-
 **Non-Normative Portions**
 
 All contents of this document are normative except portions explicitly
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,18 +67,29 @@ Note: This is an example of a non-normative note.
...
 # URL
 
-A GraphQL server operates on a single URL and all GraphQL requests for a given service should be directed
-at this URL. Other protocols may also use that URL.
+A GraphQL over HTTP compliant server MUST designate at least one URL that handles GraphQL requests.
+
+It is allowed for the same GraphQL schema to be available on multiple URLs.
+
+All GraphQL operations that are available in the schema MUST be available on each designated URL.
+That means the client can send all their GraphQL requests to a single endpoint, regardless
```
spawnia, line 84: I think this is the reasoning for serving the entire schema on a single URL, it is valuable to mention that in the spec.

ralfhandl: I'd rather first stress that a dedicated URL has to support everything in "its" schema, so clients can send all requests based on that schema to the same URL.

And then point out that there may be multiple URLs, each serving a potentially (but not necessarily) different schema.

README.md
```diff
@@ -68,18 +67,29 @@ Note: This is an example of a non-normative note.
...
+GraphQL naturally follows the HTTP request/response message model,
+providing a GraphQL request in an HTTP request and GraphQL response in an HTTP response.
 
 # URL
 
-A GraphQL server operates on a single URL and all GraphQL requests for a given service should be directed
-at this URL. Other protocols may also use that URL.
+A GraphQL over HTTP compliant server MUST designate at least one URL that handles GraphQL requests.
+
+It is allowed for the same GraphQL schema to be available on multiple URLs.
```
ghmcadams, line 81: I wonder if this line is needed (I doubt that many people have a need to run the same schema more than once on a server).  It might help to be more generic.  Something that tells people that each URL that handles GraphQL requests MUST be an independent GraphQL endpoint (even one with the same schema), as long as they are each complete.

How about:

```
Multiple URLs MAY exist on a server to handle GraphQL requests.  These URLs MUST be independent and complete.
```

README.md
```diff
@@ -68,18 +67,29 @@ Note: This is an example of a non-normative note.
...
+
+All GraphQL operations that are available in the schema MUST be available on each designated URL.
+That means the client can send all their GraphQL requests to a single endpoint, regardless
+of the contained Operations, Fragments or Variables.
+
+Other protocols may also use those URLs.
```
ghmcadams, line 87: I'm not sure `protocols` is the right word here.

README.md
spawnia: I think this is valuable to avoid ambiguity in scenarios where a server migrates its GraphQL endpoint to a different URL, offering both in a transitional period.

spawnia: Do you think the sentence above covers the completeness aspect?

> All GraphQL operations that are available in the schema MUST be available on the URLs dedicated to it.

README.md
spawnia: I expanded that bit.

README.md
spawnia: Added the first part of your suggestion.

I'd rather have the spec be a bit more wordy and possibly redundant than to be ambiguous.

README.md
spawnia: > I'd rather first stress that a dedicated URL has to support everything in "its" schema, so clients can send all requests based on that schema to the same URL.

I tweaked the wording to stress that bit.

> And then point out that there may be multiple URLs, each serving a potentially (but not necessarily) different schema.

This was also suggested by @ghmcadams, i added a note saying that multiple URLs may exist.

spawnia:
Thanks for your suggestions so far. I expanded the text to be more explicit about the aspects you mentioned.

While i do think the spec should be precise, i would rather explain things a bit more in depth than to aim for maximum conciseness.

README.md
```diff
@@ -68,18 +67,31 @@ Note: This is an example of a non-normative note.
...
-A GraphQL server operates on a single URL and all GraphQL requests for a given service should be directed
-at this URL. Other protocols may also use that URL.
+A GraphQL over HTTP compliant server MUST designate at least one URL that handles GraphQL requests.
+
+All GraphQL operations that are available in a schema MUST be available on all URLs that are designated to it.
+That means the client can send all their GraphQL requests to a single endpoint, regardless
+of the contained Operations, Fragments or Variables.
+
+Multiple URLs MAY exist on a server to handle GraphQL requests, potentially serving different
+but independent schemas. It is allowed for the same GraphQL schema to be available on multiple URLs.
```
ghmcadams, line 86:
```suggestion
but independent schemas. The same GraphQL schema MAY be available on multiple URLs on the server.
```

README.md
```diff
@@ -68,18 +67,31 @@ Note: This is an example of a non-normative note.
...
+
+Other protocols or services may also use those URLs, as long as they don't conflict with the
+server's responsibility to handle GraphQL requests.
```
ghmcadams, line 89: I believe `protocol` is not the right word here.  GraphQL is not a protocol.  HTTP is a protocol.  How about this:

```suggestion
Those URLs MAY also handle other types of requests, as long as they don't conflict with the
server's responsibility to handle GraphQL requests.
```

spawnia: Well, it means the URL may also be used to handle FTP or other protocols. Maybe this would be even for generic:

```suggestion
Those URLs MAY also be used for other purposes, as long as they don't conflict with the
server's responsibility to handle GraphQL requests.
```

ghmcadams: This wording is much better.

README.md
```diff
@@ -68,18 +67,31 @@ Note: This is an example of a non-normative note.
...
-A GraphQL server operates on a single URL and all GraphQL requests for a given service should be directed
-at this URL. Other protocols may also use that URL.
+A GraphQL over HTTP compliant server MUST designate at least one URL that handles GraphQL requests.
+
+All GraphQL operations that are available in a schema MUST be available on all URLs that are designated to it.
+That means the client can send all their GraphQL requests to a single endpoint, regardless
+of the contained Operations, Fragments or Variables.
+
+Multiple URLs MAY exist on a server to handle GraphQL requests, potentially serving different
+but independent schemas. The same GraphQL schema MAY be available on multiple URLs on the server.
```
spawnia, line 86:
```suggestion
but standalone schemas. The same GraphQL schema MAY be available on multiple URLs on the server.
```

Other ideas:
- Self-sufficient
- Complete

Maybe we don't need this section at all, but would rather specify cleanly what a single URL does.

README.md
```diff
@@ -68,18 +67,31 @@ Note: This is an example of a non-normative note.
...
+GraphQL naturally follows the HTTP request/response message model,
+providing a GraphQL request in an HTTP request and GraphQL response in an HTTP response.
 
 # URL
 
-A GraphQL server operates on a single URL and all GraphQL requests for a given service should be directed
-at this URL. Other protocols may also use that URL.
+A GraphQL over HTTP compliant server MUST designate at least one URL that handles GraphQL requests.
+
+All GraphQL operations that are available in a schema MUST be available on all URLs that are designated to it.
```
spawnia, line 81: TODO for myself: Work in the relation between schema and introspection.

spawnia, line 81:
```suggestion
All Queries and Mutations that are available in a schema MUST be available on all URLs that are designated to it.
```

sungam3r: What about subscriptions?

spawnia: Subscriptions often have some kind of redirection going on where multiple servers are involved. We might specify that in the future, but don't have a clear answer now.

spawnia:
I incorporated feedback from the comments and the last working group meeting. If there is nothing seriously wrong, we can merge this for now to allow fine-grained iteration later.

sjparsons:
I figured we can just merge for now and take any future modifications in a future PR.

PR #122, opened 2020-07-08
Converting spec to spec-md format.

mmatsa:
Converting README.md spec to spec/GraphQLOverHTTP.md in spec-md.
I tried to in general make neither content nor editorial edits.
Setting up automation of generating the HTML and pushing it to
a website is separate, therefore I left the README.md with a
reference to the new spec markdown location but did not add
a link to the HTML.
See https://github.com/APIs-guru/graphql-over-http/issues/8

@IvanGoncharov Please review, including:
- let me know what about spec-md I missed for update
- I didn't want to leave the whole copy in README.md, but I also didn't want to leave nothing there.  I don't know if it's the right balance, please let me know what you recommend
- Who else should review?

Here is the generated HTML - just remove the `.txt` and the end of the filename and view it:
[GraphQLOverHTTP.html.txt](https://github.com/APIs-guru/graphql-over-http/files/4888866/GraphQLOverHTTP.html.txt)


spawnia:
Is there a particular reason why we can not keep the spec in `README.md` for now?

mmatsa:
@spawnia - The motivation to not leave the whole thing in README.md was so that there aren't two parallel versions - which would require every PR to update both and keep both in sync.  That's similar to the main graphql spec that has the real spec in spec-md only but has a useful description of GraphQL in README.md.  In this case, the whole spec is so short that we could also just leave the single-source in README.md *only*.  I was assuming it might get longer over time and better to have the spec in `/spec/` directory and the summary in README.md.  I raised that question above about the balance here, I'm not sure what the answer should be and I'm open to any of those possibilities.
CC @sjparsons

sungam3r:
>  it might get longer over time 

it will be so

sjparsons:
I'm very much in favor of moving the spec into the `/spec` folder. I think over time it will make sense to break it up into sections -- makes PRs easier and makes management easier. I'm not too precious about what's on the front README.md. We can easily change it, so this seems like a great step.

README.md
```diff
@@ -8,188 +8,5 @@
...
-HTTP is the most common choice as the client-server protocol when using GraphQL because of its ubiquity.
-However the [GraphQL specification](https://graphql.github.io/graphql-spec/) deliberately does not specify the transport layer.
-
-The closest thing to an official specification is the article [Serving over HTTP](https://graphql.org/learn/serving-over-http/).
-Leading implementations on both client and server have mostly upheld those best practices and thus established
-a de-facto standard that is commonly used throughout the ecosystem.
-
-This specification is intended to fill this gap by specifying how GraphQL should be served over HTTP.
-The main intention of this specification is to provide interoperability between different client libraries, tools
-and server implementations.
```
sjparsons, line 22: I thinking we should keep these 3 paragraphs on the front page README.md since they give an excellent introduction into why this repo and spec exists at all.

mmatsa: Done.

mmatsa:
@sjparsons @IvanGoncharov - Simple enough, I think I addressed all those comments.  Please re-review.

IvanGoncharov:
@mmatsa Merged 🎉

PR #124, opened 2020-07-13
Add RFC for incremental delivery

robrichard:


rfcs/IncrementalDelivery.md
```diff
@@ -0,0 +1,45 @@
...
+
+---
+Content-Type: application/json; charset=utf-8
+Content-Length: 57
+
+{"data":{"test":"Hello World"},"path":[],"hasNext":false}
+
+-----
+```
+* The boundary used is `-` and is passed to the client in the http response's content-type header.
```
sungam3r, line 32: I co not see this in the example above.

jaydenseric:
Pretty interesting; I'll have to think about how this fits in with the [GraphQL multipart request spec](https://github.com/jaydenseric/graphql-multipart-request-spec) 🤔

@robrichard have you considered how they could be unified? E.g. someone using `@defer` and `@stream` directives alongside file uploads.

rfcs/IncrementalDelivery.md
robrichard: The `content-type: multipart/mixed; boundary="-"` header should be in the HTTP response headers. The example is the HTTP response body. Each part of the response body contains it's own set of headers, but the header defining the boundary needs to be on the HTTP response, not in the body.

sungam3r: it's worth mentioning

IvanGoncharov (approved):
I think it ready to be merged.
On last GraphQL WG we agreed to add support `@stream/@defer` across the stack.

rfcs/IncrementalDelivery.md
```diff
@@ -0,0 +1,45 @@
...
+
+An Incremental result is a GraphQL result that is split up into multiple payloads, allowing clients quicker access to parts of the results. Currently this is supported by the proposed `@defer` and `@stream` directives ([RFC](https://github.com/graphql/graphql-spec/blob/master/rfcs/DeferStream.md)).
+
+## `transfer-encoding: chunked`
+
+The HTTP response for an incrementally delivered response should contain the `transfer-encoding: chunked` response header. Chunked transfer encoding allows the body of the response to be delivered as a series of chunks, allowing clients to read each chunk of the response as it is sent without waiting for the entire response.
+
+## `content-type: multipart/mixed`
+
+The HTTP response for an incrementally delivered response should contain the `content-type: multipart/mixed; boundary="-"` response header and conform to the [specification of multipart content defined by the W3 in rfc1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). An example response body will look like:
```
benjie, line 15: Must the boundary be `"-"` specifically? Could other boundaries be used?

benjie (approved):

rfcs/IncrementalDelivery.md
robrichard: I think we should support arbitrary boundaries. [fetch-multipart-graphql](https://github.com/relay-tools/fetch-multipart-graphql) currently only supports `"-"` but I will update that and this document to allow any boundary.

IvanGoncharov: @benjie @robrichard Sorry, I didn't notice this comment before merging.
On the last "GraphQL over HTTP" WG we decided on merging and just forgot to actually do it.
Hope it didn't disrupt your workflow and @benjie suggestion can be addressed as separate PR.

benjie: I've filed an issue here: https://github.com/graphql/graphql-over-http/issues/135 (@IvanGoncharov if you could assign it to @robrichard that would be helpful).

acao:
@jaydenseric to clarify, i think what @robrichard is saying is that non-file fields can use `@defer` and `@stream` with multipart alongside your multipart file spec. in slack today @robrichard confirmed internally they have a client that is successfully able to do as such.

@jaydenseric speaking of which, would love to help advance file spec, and maybe we can revisit how `@defer` and `@stream` impact your spec if at all, and revise them if/as needed? I'm going to open an issue in your spec repo to discuss this further!

PR #125, opened 2020-07-17
Recommend concrete status codes for specific error cases

spawnia:
This PR builds upon https://github.com/graphql/graphql-over-http/pull/79 and adds additional specifications atop.


spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,38 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
-If the operation failed before or during execution, due to a syntax error, missing information, validation error
-or any other reason, then response should use the 4XX or 5XX status codes.
-It is recommended to use the same error codes as the [reference implementation](https://github.com/graphql/express-graphql).
+If the response has Content-Type GraphQL and contains a non-null `data` entry,
+then it MUST have status code `2xx`, and it SHOULD have status code `200`.
+
+If the response has Content-Type GraphQL and has status code `2xx`,
+the entry `data` entry must be either:
+- equal to `null` (in case of an execution error)
+- not present (in case of a validation error)
```
benjie, line 210: This is wrong, these rules apply to non-`2xx` status codes. Suggest rewording to:

```suggestion
If the response has Content-Type GraphQL and has a non-success
status code (i.e. `4xx` or `5xx`), the entry `data` entry must be
either:
- equal to `null` (in case of an execution error)
- not present (in case of a validation error)
```

spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,38 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
+There are four types of validation errors:
+
+1. Unparseable or invalid request body, for example `NONSENSE` or `{"qeury": "{__typename}"}`
+2. GraphQL Specification document validation errors (valid syntax, no loops in fragments, etc)
+3. Custom document validations that can be performed before execution (e.g. graphql-depth-limit)
+4. Runtime validations performed by resolvers (i.e. during the execution of the GraphQL operation).
+
+The error types 1. and 2. prevent execution of the request entirely and MUST result in
+status code `400` (Bad Request). Types 3. and 4. may still allow partial execution and
+MUST result in status code `200`.
```
benjie, line 228: This currently disagrees with the flexibility of the wording above. I've softened it:

```suggestion
Error types 1. and 2. prevent execution of the request in its entirety; they MUST result in
status code `4xx` or `5xx`, and SHOULD result in status code `400` (Bad Request). Types 3. and 4. may still allow partial execution; they
MUST result in status code `2xx`, and SHOULD result in status code `200` (Okay).
```

spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,38 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
+
+If an unauthenticated client is not allowed to access the schema, the server MUST respond with
+status code `401` (Unauthorized).
```
benjie, line 231:
```suggestion
If a client is not allowed to access the schema, the server MUST respond with
status code `4xx`, SHOULD respond with `401` (Unauthorized) or `403` (Forbidden) as appropriate, and MAY respond with status code `404` (Not Found) if security concerns demand it.
```

spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,38 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
+
+If the server failed unexpectedly, it MUST respond with status code `500` (Internal Server Error).
```
benjie, line 233: Limiting this to 500 seems unnecessary; timeouts, bad gateways, downtime, etc are all reasonable status codes to use here.

```suggestion
If the server failed unexpectedly, it MUST respond with the appropriate `5xx` status code.
```

spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,38 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
+
+If a client is not authorized to access any of the requested fields, the server MUST respond with
+status code `403` (Forbidden) and not return any `data`.
```
benjie, line 236: I think this sentence should be deleted in its entirety.

```suggestion
```

spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,38 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
+If only some requested fields are restricted, and the server can provide at least partial `data`,
+the response MUST have status code `200` (OK) and include `errors` as part of the GraphQL response.
```
benjie, line 238: This should be a non-normative note, as really this is just a description of a partial success. Alternatively remove it.

```suggestion
> NOTE: if certain fields within the GraphQL schema are restricted such that certain clients are not permitted to execute them, access to these fields should result in standard GraphQL field errors following the [error bubbling logic in the GraphQL spec](https://spec.graphql.org/June2018/#sec-Errors-and-Non-Nullability), resulting in a partial-success. As outlined above, this should result in partial `data`, a list of `errors`, and a `2xx` status code, preferably `200` (Okay).
```

spec/GraphQLOverHTTP.md
spawnia: This section is also in https://github.com/graphql/graphql-over-http/pull/79, which this PR depends upon.

spawnia: [`9e75abc` (#79)](https://github.com/graphql/graphql-over-http/pull/79/commits/9e75abccfb0030dbbddf324b5fa621bf37b8e919)

spec/GraphQLOverHTTP.md
jayadeba: Mutations are usually atomic operations so the case of partial execution doesn't exist (except for say chained mutations).  How about leaving it to the application's business logic to determine the best HTTP error code suitable for the use case, without being prescriptive about which exact error code should it throw. For example, incase of idempotent mutation, for conflict scenarios you throw 409. downstream connection failure errors while executing a mutation can throw 500 and the upstream client can retry the request using exponential backoff.

benjie: The selection set on a mutation field is effectively a query, so partial execution applies to mutation operations in just the same way it does to query operations.

Further, mutation operations can contain more than one mutation (field) within them, so using a code such as 409 does not make sense - which of the mutations does it apply to? If one succeeded and one failed, do we use the 409 for the failure or the 200 for the success?

```graphql
# Assuming this results in a 409 status; which one of these went wrong? Did they both fail?
mutation Conflict {
  createUser(username: "Bob") { id name }
  updateUser(username: "Alice", toUsername: "Alan") { id name }
}
```

We should do our best to help people avoid these ambiguities, that's our job writing the spec - to make sure that all situations are handled and clear. This is why we've stated that as soon as execution is successfully invoked, the response should be a 200 status (successful GraphQL execution) and it is then up to the client to process the results of the execution which it is best placed to do, given it understands GraphQL and knows what request it sent.

jayadeba: I think we can separate a purely atomic mutation operation (that can either fail or pass- almost all mutations fall in this category) from a composite mutation operation (which involves multiple steps) or a chained mutation like the example you gave above (composite or chained mutations are used only in special cases).  The idempotent mutation example that I gave above is an atomic mutation (for example `makePayment`) and considering that `makePayment` is declared idempotent, when `makePayment` is already processing and another duplicate request arrives, the error that  is appropriate to throw is a 409 Conflict. Throwing a 200 here doesn't help. Same with 500 errors when there's a downstream connection failure/timeout, while processing the request during the graphql execution). The intermediary proxies, monitoring systems in any enterprise infrastructure won't introspect the response body to log the errors, create appropriate alerts or do automatic retries in case of 500 errors and so on. Throwing an HTTP error code as-is from atomic mutation operation will make the existing web infra work as-is. Neither the API provider nor the API consumer need to write any custom logic for understanding GraphQL responses. This is the single most complaint (GraphQL error handling) I hear from most clients on why they don't want to consume or build **public** GraphQL APIs. I think we have an opportunity to address that gap through this working group. Solving this problem will help towards more providers building GraphQL public APIs.

My recommendation (If these make sense) :

1- We allow the API provider to emit one of the standard HTTP status codes as appropriate without being prescriptive about which code to use when or instead of saying use 200 only if the error occurs during graphql execution. Instead we let the application decide on which status code to emit (e.g use 200 for chained mutations if there are partial errors or appropriate 4XX/5XX when there= are errors, while executing atomic mutations), regardless of whether the error occurred prior to execution or during execution.

2- We should think this specification as the HTTP binding spec for GraphQL, which means, we should let the HTTP protocol dictates the rules. That said we should keep the core GraphQL constructs as-is (of course we shouldn't allow someone to use a PUT method for a mutation:-)) . As an example from another spec we can check the [CNCF Cloudevent Specification](https://github.com/cloudevents/spec). The Cloudevent envelope format is a protocol agnostic envelope to describe events and CNCF defines binding rules for each type of transport (Including HTTP). See all binding specs [here](https://github.com/cloudevents/spec)  and let the protocol dictate the status code, security etc.

3- We should define some http specific directives to define HTTP metadata (like request/response headers, declaring a method is idempotent, status codes, produces/consumes etc.

benjie: To summarize, I think you're saying:

If:

- the operation is a mutation, and
- the operation contains exactly one non-introspection (i.e. not `__typename`) root-level field, and
- that field throws an error

Then:

- the server MAY choose to throw a non-200 status code

Else:

- the server MUST result in status code `2xx`, and SHOULD result in status code `200` (Okay)`

Your example of 500 errors getting auto-retried on the back-end is where I've drawn these rules from; for example if the field `name` in my `Conflict` example above were to fail due to network connectivity (but all other fields, including the mutation itself, succeeded) then presumably this should not auto-retry since it's an expected failure state that GraphQL was designed to handle gracefully with its nullable-by-default partial-response + errors.

If this is an inaccurate description of your position, please could you expand exactly when you think non-200 status codes should be valid?

---

Separately, how do you intend to indicate that a specific mutation field is idempotent, and which part of the stack is the consumer of that information?

jayadeba: @benjie almost. Let me add some examples for atomic mutations and illustrate the scenarios where the client always executes one mutation at a time

**Make Payment**:

~~~
 {
     "query" : "mutation makePayment (...)"
     "variables" : {...}

}
~~~

**Create User**:

~~~
 {
     "query" : "mutation createUser (...)"
     "variables" : {...}

}
~~~

In the examples above;

* Both the Mutations `makePayment` and `createUser` are atomic and do only one atomic action.
* Both of the above mutations are always executed independently but never together
* There's no chance of partial failure here. Any of the mutations when executed either fail or succeed.


Considering the above, here are the error scenarios;
* `makePayment` can return 400, when the receiver's emai ld/phone number is incorrect or account number is invalid or aaccount is inactive and payment cant be made.

* `makePayment` is idempotent (implements exactly once semantics) so when you send a duplicate `makePayment` call with same payload again, the server would reject the call with a 409 Conflict error.

* When `makePayment` cant reach the card network server (Visa or Master card) that the sender used the card to make payment or the request with card network times out, `makePayment` would like to return a 500/504 error, as its an intermittent recoverable error and the likelihood of payment being successful if it is retried again is very high.Returning a 500 to the API Gateway or API client, will make them retry the request. That is how all intermediary proxies are configured to do automatic retries on 500s using exponential backoff.

* `makePayment` may return a 403 Authorization error if the user is not allowed to make payments (say its a child account and parents have restricted the child account permission to only receive payments but not make any payments)

*A monitoring system will send alerts to customer support team if there are too many payment failures. Monitoring systems look for HTTP error codes. Similarly a monitoring system will report to security incident team if tehre are too many authorization errors (say 403). 

* A rate limiter system may block requests if there are too many requests from a device that fail with authorization errors (Rate limiter systems depend on HTTP status codes).

So all atomic operations should be able to return 4XX/5XX errors as appropriate and that makes the errors explicit and clear and uses HTTP protocol fully. Any traditional client with standard HTTP infrastructure doesn't need to write any custom error handling code to integrate with GraphQL.

**The other Use cases (which result in Partial failures so 200 response is OK):**
In the case of chained mutations (your example above, there are always chances of partial errors so a 200 response with `data` and `errors` are appropriate.

To conclude, my recommendation is that we allow the GraphQL service implementations return appropriate HTTP status codes, without being prescriptive about when to throw 4XX/5XX vs when to throw 2XX. 

Rewarding idempotency, you can for example, come up with a directive such as `@idempotent`. When you declare a method idempotent it helps client to retry such method automatically in case of request timeouts. on the server side, its natural that the implementation enforces a constraint wherein all clients need to send an unique idempotency key for every mutation call that is declared idempotent (this is a standard industry pattern). If the client doesnt send a idempotency-key in the mutation call, the server rejects the request with a 400 error.

Example:

~~~
type mutation {

    makePayment(...) @idempotent

}
~~~

Hope this helps.

benjie: > Both of the above mutations are always executed independently but never together

This seems to be a human rule rather than a computer rule. How can the computer know this? How would it know the following operation is forbidden?

```graphql
mutation TakePaymentAndCreateAccount ($accountDetails: CreateAccountInput!, $paymentDetails: TakePaymentInput!) {
  takePayment(input: $paymentDetails) { success errors { code message } }
  createAccount(input: $accountDetails) { userId }
}
```

It's very common during a checkout process for the website to also invite you to create an account. How can we expand this to GraphQL in general that's not specific to your server's setup?

---

Regarding your example of the `@idempotent` directive on a mutation, there's currently no way for baseline Spec compliant GraphQL servers to indicate that a mutation has this tag; it's something that I hope we will have in the coming years but work on that is ongoing. So the server cannot indicate this; and further even if the server _did_ indicate this _through GraphQL_ then for an intermediary service to understand it, it would need to understand GraphQL, at which point using the HTTP status code for auto-retry becomes unnecessary since the gateway can extract the information it needs from the GraphQL request directly (although slight rewriting of the query might be necessary if the user did not request sufficient information). The same "proxy server needs to understand GraphQL" applies also to the case of a client-indicated `@idempotent` directive which is what your example demonstrates (although I'm not sure if it is what you meant?).

benjie: (I note "baseline Spec compliant GraphQL servers" because some implementations (GraphQL Java?) have experimental support for indicating directives applied to fields within the introspection results.)

spawnia: > There's no chance of partial failure here. Any of the mutations when executed either fail or succeed.

The fact that mutations also return a selection set makes this problematic. Let's expand upon the example query:

```graphql
mutation {
  createAccount(input: $accountDetails) {
    id
    friends {
      someAccountDetailOnlyAvailableToAdmins
    }
  }
}
```

Suppose you were to return an error code 403; how can the client differentiate between:
- they are not allowed to create accounts
- they are not an admin and cannot view certain fields

A single status code provides too little information here. Given the complex nature of queries, reducing the result down to a single status code simply is not feasible once execution started. The HTTP semantics for looking at the status codes of individual requests simply falls apart, since it is based upon the presumption that *one request does one thing and one thing only*.

GraphQL does provide means of communicating errors within the execution within the result. That is by design. I get that adopting such a paradigm shift may be challenging and hinder adoption, but shoehorning HTTP semantics on top will not help in the long run.

spawnia: @benjie thanks for the suggestion, I added it in but will leave this discussion open for now.

spec/GraphQLOverHTTP.md
spawnia: That's a great way of putting it and should be sufficiently flexible.

spec/GraphQLOverHTTP.md
spawnia: Thinking about this more, I have come around to agree with you on this. Some fields may error out because of missing authentication, while others may also error out because of other kinds of errors. Rubber-stamping that with a `403` is going too far.

spec/GraphQLOverHTTP.md
spawnia: This was added in contrast to the above statement about returning `403` if all fields are restricted, which I now removed as of your suggestion. Given that the GraphQL spec itself already describes partial success and we mention it in a section above, I will just remove it.

spec/GraphQLOverHTTP.md
```diff
@@ -201,8 +201,35 @@ may not be the GraphQL server but instead some intermediary such as API gateways
...
+4. Runtime validations performed by resolvers (e.g. during the execution of the GraphQL operation).
+
+Error types 1. and 2. prevent execution of the request in its entirety; they MUST result in
+status code `4xx` or `5xx`, and SHOULD result in status code `400` (Bad Request).
+Types 3. and 4. may still allow partial execution; they MUST result in
+status code `2xx`, and SHOULD result in status code `200` (Okay).
+
+If a client is not allowed to access the schema, the server MUST respond with
+status code `4xx`, SHOULD respond with `401` (Unauthorized) or `403` (Forbidden) as appropriate,
+and MAY respond with status code `404` (Not Found) if security concerns demand it.
```
ghmcadams, line 233: Is this needed (it is already covered in `MUST respond with
status code '4xx'`)?

If you feel it is still needed, then I would prefer that `if security concerns demand it.` is removed (motivations are unenforceable).

spawnia: I shortened that part and just prescribe a `4xx`.

spawnia:
I rewrote the list of validation error cases into a more loose collection of example cases (sorry @benjie 😉). The previously used numbered list format made it seem like it was exhaustive, which is pretty hard to do.

spec/GraphQLOverHTTP.md
```diff
@@ -215,3 +215,31 @@ as long as the server is still able to produce a well-formed response.
...
 In case of errors that completely prevent the successful execution of the request,
 the server SHOULD respond with the appropriate status code depending on the concrete error condition.
+
+The following examples provide guidance on how to deal with specific error cases:
+
+### Unparseable or invalid request body
+
+For example: `NONSENSE`, `{"qeury": "{__typena`
+
+Completely prevents execution of the GraphQL operation and SHOULD result in status code `4ß0` (Bad Request).
```
ghmcadams, line 225: There seems to be a unicode problem here (`4ß0`)

spec/GraphQLOverHTTP.md
```diff
@@ -215,3 +215,31 @@ as long as the server is still able to produce a well-formed response.
...
+- custom validation, for example: depth limit, complexity limit
+
+The server MAY choose to:
+- deny execution, the status code SHOULD be `400` (Bad Request)
+- attempt execution, the status code will depend on the validity of the result
+
+### Runtime validation
+
+Validation steps performed by resolvers during execution of the GraphQL operation.
+The status code will depend on the validity of the result.
```
ghmcadams, line 240: In previous conversations, it was said that any validation done INSIDE the resolvers should result in a `2xx` (preferably a `200`) status code response.  Individual resolvers could have code called that is validations but are in reality arbitrary.

mike-marcacci:
Hey folks! I've been crazy busy elsewhere and have only been able to follow the evolution here very lightly. This just hit my inbox and I happened to notice that this language would conflict with HTTP optimization mechanisms that would return `304` if the resulting response body matched the `etag` header specified in the request. This isn't something I anticipate being particularly useful here, but many HTTP servers and proxies will automatically handle this, and it should be clear that a `304` is equivalent to a 200.

This probably isn't the optimal place for me to add this comment, but I figured I'd throw it out anyhow, before I get pulled away 🙃

spec/GraphQLOverHTTP.md
mmatsa: That's a great point - but I'd rather stick to the idea of "any time a partial result is returned (non-null data) then it must be 2xx and not 4xx.  I don't think the spec should explicitly make requirements on the opaque parts of the server - if the client can't tell whether something happened inside or outside of resolvers then I don't think the spec should make assertions about it.  I'm not sure "non-null data" is enough, but maybe it's a starting point?

mmatsa:
@mike-marcacci - that's a great comment, we want to take things like that into account, but I'm not sure which part of this PR is a conflict for a `304` response.  Can you say which part you noticed?

> Note: For any non-`2XX` response, the client should not rely on the body to be in GraphQL format

That seems to apply well to `304`, since I think it has no body.

> If the response has Content-Type GraphQL and contains a non-null `data` entry, then it MUST have status code `2xx`

I assume a `304` would not have a body and therefore this clause doesn't apply.

And the failing-validation and non-allowed sections don't seem to apply.

I'm not an expert at this, so I know I could be missing something big - can you explain which part of the PR and why it's a problem?

- Any validation

spec/GraphQLOverHTTP.md
```diff
@@ -215,3 +215,31 @@ as long as the server is still able to produce a well-formed response.
...
+
+### Document validation
+
+Includes validation steps that run before execution of the GraphQL operation:
+- [GraphQL specification validation](http://spec.graphql.org/June2018/#sec-Validation)
+- custom validation, for example: depth limit, complexity limit
+
+The server MAY choose to:
+- deny execution, the status code SHOULD be `400` (Bad Request)
+- attempt execution, the status code will depend on the validity of the result
```
benjie, line 235: Why are we allowing the server to attempt execution when validation fails? Surely that could lead to a whole host of issues (for just one example, infinite recursion from recursive fragments might be permitted if just that validation rule failed). Is it specific to custom validation rules?

I'm not saying we shouldn't (though that is the direction I'm leaning); but I'd like to understand the reasons why we think we should (and I'd like that added as a non-normative note).

benjie (changes requested):

spec/GraphQLOverHTTP.md
mmatsa: @benjie - Let me know if I'm missing your point.  I certainly support a non-normative note, but I don't think it's the job of the `GraphQL over HTTP` spec to allow or not allow execution on an invalid query.  The `GraphQL` spec itself says:

> An invalid request is still technically executable, and will always produce a stable result as defined by the algorithms in the Execution section, however that result may be ambiguous, surprising, or unexpected relative to a request containing validation errors, so execution should only occur for valid requests.

I read that as "you *should* only execute valid queries, but you *may* execute invalid queries".  I might be missing somewhere else in the spec that says differently?  If I'm correct and if every legal GraphQL server may be used over `GraphQL over HTTP`, then we should be controlling how they reply over HTTP whether or not they choose to execute an invalid query, exactly as long as the spec allows it.

While I generally always support vaIidation before execution, in the "real world" I think it's reasonable to not validate queries that you know to be valid - for example if they are generated by your code or already validated by other middleware - and then what if there's a bug and it wasn't valid after all?  I think that our spec should be resilient enough to explain simple rules that are still relevant like "any non-null data must have a 2xx status code" etc.

Even if you agree with that, I would definitely support rewriting the text to be clearer what we're saying and not imply that we recommend executing invalid queries.

benjie: I think the first bullet should be SHOULD rather than MAY. At the moment it sounds like they’re both equivalent.

Though the GraphQL spec does indeed talk about what is _technically_ possible (this is necessary to help library authors structure their GraphQL libraries), I do think the GraphQL over HTTP Spec should be recommending what users should do when exposing a GraphQL schema over HTTP, and validation is essential to that (whether it’s done when the request is received, or beforehand e.g. during a build phase with persisted operations, or cached from the last time they saw that same document).

mmatsa: I think I'm happy with SHOULD there.  I totally agree that they're not equal clauses.

Brainstorming - Might it be better instead to rewrite to avoid mentioning it?  I think our spec might be stronger if it only mentions internals for the GraphQL execution engine in non-normative notes to help implementers, but for the requirements sticks to things like "a non-null `data` MUST have a 2xx, and SHOULD have a 200."  (Then leaving the internals of what execution engine paths yield that non-null `data` in the first place to the GraphQL spec as well as the non-normative notes/section.)

For your other point, I agree that validation is essential - often it's in one of the ways you mention but sometimes at our customers it's done at a different point in the network topology and the current server has a "guarantee" that the incoming transactions are valid (internal generated queries, or already validated in the DMZ, etc.), and obviously sometimes those "guarantees" fail.  That's not universal, and not limited to GraphQL, but I think it's a real tradeoff out there.

mmatsa: I just re-checked the PR and I see two things I'd forgotten (sorry I didn't check earlier):
- 1. this is already inside an 'example' and thus I assume this whole part of the text is non-normatime.  As such, I'm happy for it to say SHOULD, but I think it could be stronger, and say something like "the server would probably return a 4xx" or something like that and not even mention the other option.
- 2. it explicitly lists multiple types of "bad request", one being a strict spec failure to validate and the other being "custom validation, for example: depth limit, complexity limit" - for these kinds of 'extra' checks I think that both "log a warning and continue" and "error immediately" are acceptable options.  I don't know who originally wrote that text but maybe that was the motivation.

spec/GraphQLOverHTTP.md
ghmcadams:
```suggestion
Completely prevents execution of the GraphQL operation and SHOULD result in status code `400` (Bad Request).
```

spec/GraphQLOverHTTP.md
ghmcadams: Isn't it true that once any resolver executes, it guarantees a non-null data in the response?  If so, than the status code MUST be 2xx.

```suggestion
The status code (as mentioned above) MUST be 2xx, and SHOULD be 200 (Okay).


```

spec/GraphQLOverHTTP.md
ghmcadams, line 235:
```suggestion
The server SHOULD deny execution with a status code of `400` (Bad Request).  An invalid request is still technically executable, therefore the server MAY attempt execution.  In this event, the status code will vary based on other parts of this document.
```

spawnia: Thanks, incorporated parts of this into an updated version.

spawnia: Thank you for the feedback, I think it is generally a good recommendation to abort. I came up with a use case for continuing regardless: persisted queries that were previously known to be valid. I adapted this section accordingly: https://github.com/graphql/graphql-over-http/pull/125/commits/5520f071f64c9fad6145bc3f49b1e16504f873b0

spec/GraphQLOverHTTP.md
spawnia: It could be the case that a root resolves for a non-nullable field aborts because of a validation error. In that case `data` could be forced to become `null`, thus allowing (but not forcing) a non-2xx status code.

Since this section is of the spec is us giving recommendations for specific scenarios, I changed this to encourage returning `200` for consistency across fields: [`663ef69` (#125)](https://github.com/graphql/graphql-over-http/pull/125/commits/663ef69197e33c5625816b398366df65c7230837)

spawnia:
Thanks @ghmcadams for reviving the discussion here and reminding me, I addressed the remaining discussions.

ghmcadams (approved):

PR #137, opened 2020-09-07
Minor tweaks, explicitly call out the link to HTTP safe methods

kittylyst:


spec/GraphQLOverHTTP.md
```diff
@@ -170,9 +170,14 @@ http://example.com/graphql?query=query(%24id%3A%20ID!)%7Buser(id%3A%24id)%7Bname
 
 Note: {query} and {operationName} parameters are encoded as raw strings in the query component. Therefore if the query string contained `operationName=null` then it should be interpreted as the {operationName} being the string `"null"`. If a literal `null` is desired, the parameter (e.g. {operationName}) should be omitted.
 
+GET requests MUST only be used for executing queries. If the values of {query} and {operationName} indicates that a non-query operation is to be executed, the server SHOULD immediately respond with error status code 405 Method Not Allowed and halt execution.
```
spawnia, line 173: Can you factor out this part into a separate PR? That would allow us to get the minor edits through the door faster and have a focused discussion on this substantial change.

kittylyst: Sure. Other PR to follow.

spawnia (approved):

sungam3r (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -138,9 +140,9 @@ Note: {variables} and {extensions}, if set, must have a map as its value.
 
 ## GET
 
-For HTTP GET requests, the query parameters should be provided in the query component of the request URL in the form of
+For HTTP GET requests, the query parameters MUST be provided in the query component of the request URL in the form of
 `key=value` pairs with `&` symbol as a separator and both the key and value should have their "reserved" characters percent-encoded as specified in [section 2 of RFC3986](https://tools.ietf.org/html/rfc3986#section-2).
-The unencoded value of the {variables} parameter should be represented as a JSON-encoded string.
+The unencoded value of the {variables} parameter SHOULD be represented as a JSON-encoded string.
```
benjie, line 145: This should probably be a `MUST`
```suggestion
The unencoded value of the {variables} parameter MUST be represented as a JSON-encoded string.
```

spec/GraphQLOverHTTP.md
```diff
@@ -172,7 +174,7 @@ Note: {query} and {operationName} parameters are encoded as raw strings in the q
 
 ## POST
 
-A standard GraphQL POST request should have a body which contains values of the request parameters encoded according to
+A standard GraphQL POST request SHOULD have a body which contains values of the request parameters encoded according to
```
benjie, line 177:
```suggestion
A standard GraphQL POST request MUST have a body which contains values of the request parameters encoded according to
```

spec/GraphQLOverHTTP.md
```diff
@@ -188,21 +190,21 @@ Note: both query and mutation operations can be sent over POST requests.
...
 
-When a GraphQL server receives a request, it must return a well‐formed response. The server’s
+When a GraphQL server receives a request, it MUST return a well‐formed response. The server’s
 response describes the result of executing the requested operation if successful, and describes
 any errors encountered during the request.
 
 ## Body
 
-If the server's response contains a body it should follow the requirements for [GraphQL response](https://graphql.github.io/graphql-spec/June2018/#sec-Response).
+If the server's response contains a body it SHOULD follow the requirements for [GraphQL response](https://graphql.github.io/graphql-spec/June2018/#sec-Response).
```
benjie, line 199:
```suggestion
If the server's response contains a body it MUST follow the requirements for [GraphQL response](https://graphql.github.io/graphql-spec/June2018/#sec-Response).
```

spec/GraphQLOverHTTP.md
```diff
@@ -188,21 +190,21 @@ Note: both query and mutation operations can be sent over POST requests.
...
 
-Note: For any non-2XX response, the client should not rely on the body to be in GraphQL format since the source of the response
+Note: For any non-2XX response, the client SHOULD NOT rely on the body to be in GraphQL format since the source of the response
 may not be the GraphQL server but instead some intermediary such as API gateways, firewalls, etc.
 
 ## Status Codes
 
 If the `data` entry in the response has any value other than `null` (when the operation has successfully executed
-without error) then the response should use the 200 (OK) status code.
+without error) then the response SHOULD use the 200 (OK) status code.
```
benjie, line 207:
```suggestion
without error) then the response MUST use a 2xx (success) status code, and SHOULD use the 200 (OK) status code.
```

spec/GraphQLOverHTTP.md
```diff
@@ -188,21 +190,21 @@ Note: both query and mutation operations can be sent over POST requests.
...
 If the operation failed before or during execution, due to a syntax error, missing information, validation error
-or any other reason, then response should use the 4XX or 5XX status codes.
+or any other reason, the response SHOULD use the 4XX or 5XX status codes.
```
benjie, line 209:
```suggestion
or any other reason, the response MUST use a 4XX or 5XX status code.
```

benjie (changes requested):
Whilst formalizing this language we should be careful of our use of `SHOULD` vs `MUST`; in general we should try and use MUST to be as prescriptive as possible to help interoperatbility. I've corrected the text in line with what I believe we've discussed during the working groups.

spec/GraphQLOverHTTP.md
benjie: (Without this, client interop is going to be an issue: what formats does the server support? What formats does the client support?)

benjie:
I believe my suggestions still stand :+1:

spec/GraphQLOverHTTP.md
ghmcadams: I'm not sure this is a `MUST`.  The key words to me are `before or during execution`.  If there is a validation error on one of the fields' input arguments, isn't a 2xx response still valid?

spawnia: The current spec has already been updated here, see https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#status-codes

This PR should be rebased.

ghmcadams: Agreed

benjie: >  If there is a validation error on one of the fields' input arguments, isn't a 2xx response still valid?

If it's a GraphQL validation error it occurs before execution so must result in a 4XX/5XX error - the request is invalid and cannot be executed.

If it's an error thrown by a resolver, then it should result in something like:

```graphql
type Mutation {
  myMutation(a: Int): Int
}

mutation ThrowsError {
  myMutation(a: -3)
}
```

```json
{
  "data": {"myMutation": null},
  "errors": [
    ...
  ]
}
```

This should be 200, since "data" is present and non-null.

However in the case that you've defined the mutation such that the field is non-nullable:

```graphql
type Mutation {
  myMutation(a: Int): Int!
}
```

and then the field throws, then the result will be:

```json
{
  "data": null,
  "errors": [
    ...
  ]
}
```

This has been controversial how to handle in the past. There are arguments for both sides. The [above linked document](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#status-codes) currently states:

- present and non-null "data" -> 2XX (preferably 200)
- non-2XX -> "data" must be null/not present

It leaves the case of `"data": null` undefined - the text allows for it to be 2XX or non-2XX. Personally I think it should be a 5XX since it's a schema design issue. Take for example this mutation operation:

```graphql
mutation DoesThreeThings {
  createUser(name: "Benjie") { id name }
  orderProduct(id: 27) { deliveryDate }
  myMutation(a: -3)
}
```

If `myMutation` throws (and is non-nullable), the result will be:

```json
{
  "data": null,
  "errors": [
    ...
  ]
}
```

The user cannot know if createUser/orderProduct succeeded (though they can determine that they did not throw errors) due to a schema design fault.

mmatsa: @benjie - thanks for the detailed explanation and examples.
If we use your proposal, would we be able to go further and for purposes of HTTP status codes unify no-data and null-data as the same category?
ie:
- present and non-null `data` iff 2xx
- missing or null `data` iff non-2xx

benjie: I have a vague feeling that’s how express GraphQL works already. I would be happy with this decision, but it should be noted these two situations come from separate phases. I feel like data:null is always a server/schema error, so 5XX.

ghmcadams:
@kittylyst Please rebase this PR

PR #138, opened 2020-09-07
Establish error code 405 for attempted unsafe operation via GET

kittylyst:


spec/GraphQLOverHTTP.md
```diff
@@ -170,6 +168,11 @@ http://example.com/graphql?query=query(%24id%3A%20ID!)%7Buser(id%3A%24id)%7Bname
 
 Note: {query} and {operationName} parameters are encoded as raw strings in the query component. Therefore if the query string contained `operationName=null` then it should be interpreted as the {operationName} being the string `"null"`. If a literal `null` is desired, the parameter (e.g. {operationName}) should be omitted.
 
+GET requests MUST only be used for executing queries. If the values of {query} and {operationName} indicates that a non-query operation is to be executed, the server SHOULD immediately respond with error status code 405 Method Not Allowed and halt execution.
+
+This restriction is necessary to conform with the long-established semantics of safe methods within HTTP.
```
benjie, line 173: I agree with this change for mutations, but for subscriptions it's less clear. E.g. it would be absolutely fine to do GraphQL subscriptions over server-sent events or [comet](https://en.wikipedia.org/wiki/Comet_(programming)).

I definitely support a change that explicitly says that if the operation is `mutation` then a `GET` MUST result in a 405.

benjie (changes requested):
Thanks for this change, I think it is definitely heading the the right direction.

spec/GraphQLOverHTTP.md
kittylyst: I think I agree that it should be fine from a correctness point of view to do subscriptions over comet-style connections.

My concern, though, is that by using a GET method to trigger a subscription, we could be opening the door to inadvertent resource draining of the server by naive intermediate HTTP components which follow the letter of the spec and retry a GET call without any intervention, thus causing duplicate subscriptions. It's been a long while since I looked at this in detail, but I recall that this was a concern for early pre-WebSocket comet libraries.

benjie: Are you aware of similar issues with server-sent events?

https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

ghmcadams:
There are some unaddressed comments on this PR and some significant time has passed.  This PR will be merged soon.  Please comment as desired.

spec/GraphQLOverHTTP.md
ghmcadams: @kittylyst any thoughts on this?

ghmcadams: Looking for a resolution on this.  @benjie @kittylyst?

benjie: We could make it acceptable by saying "GET requests must not be used for mutations" rather than "only for queries".

ghmcadams: @benjie It seems we lost @kittylyst.  If you make a suggestion, I will apply it and we can merge.

kittylyst: Sorry - got dragged onto other things. "not for mutations" seems like a reasonable plan to me.

ghmcadams: Glad to have you back, @kittylyst.  Could you please make this change?

benjie, line 173:
```suggestion
GET requests MUST NOT be used for executing mutation operations. If the values of {query} and {operationName} indicates that a mutation operation is to be executed, the server SHOULD immediately respond with error status code 405 Method Not Allowed and halt execution. This restriction is necessary to conform with the long-established semantics of safe methods within HTTP.
```

benjie (changes requested):

benjie (approved):

spawnia (approved):

PR #143, opened 2020-09-23
Allow arbitrary boundaries in IncrementalDelivery RFC

robrichard:
Fixes #143

### DIFF
--- a/rfcs/IncrementalDelivery.md
+++ b/rfcs/IncrementalDelivery.md
@@ -12,7 +12,9 @@ The HTTP response for an incrementally delivered response should contain the `tr
 
 ## `content-type: multipart/mixed`
 
-The HTTP response for an incrementally delivered response should contain the `content-type: multipart/mixed; boundary="-"` response header and conform to the [specification of multipart content defined by the W3 in rfc1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). An example response body will look like:
+The HTTP response for an incrementally delivered response should conform to the [specification of multipart content defined by the W3 in rfc1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). The HTTP response must contain the `Content-Type` response header with a specified boundary, for example `content-type: multipart/mixed; boundary="-"`. Using a simple boundary of `-` works as there is no possiblity of conflict with JSON data. However, any boundary may be used.
+
+An example response body will look like:
 
 ```
 ---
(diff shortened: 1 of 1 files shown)


maraisr:
This totally makes sense! Was actually reading the spec lastnight, and was going to open an issue asking why boundaries aren't supported, then I found this! 

Great work @robrichard, and thanks for pushing this feature!

rfcs/IncrementalDelivery.md
```diff
@@ -12,7 +12,9 @@ The HTTP response for an incrementally delivered response should contain the `tr
 
 ## `content-type: multipart/mixed`
 
-The HTTP response for an incrementally delivered response should contain the `content-type: multipart/mixed; boundary="-"` response header and conform to the [specification of multipart content defined by the W3 in rfc1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). An example response body will look like:
+The HTTP response for an incrementally delivered response should conform to the [specification of multipart content defined by the W3 in rfc1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). The HTTP response must contain the `Content-Type` response header with a specified boundary, for example `content-type: multipart/mixed; boundary="-"`. A simple boundary of `-` can be used as there is no possiblity of conflict with JSON data. However, any arbitrary boundary may be used.
```
benjie, line 15:
```suggestion
The HTTP response for an incrementally delivered response should conform to the [specification of multipart content defined by the W3 in rfc1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). The HTTP response must contain the `Content-Type` response header with a specified boundary, for example `Content-Type: multipart/mixed; boundary="-"`. A simple boundary of `-` can be used as there is no possiblity of conflict with JSON data. However, any arbitrary boundary may be used.
```

rfcs/IncrementalDelivery.md
```diff
@@ -29,7 +31,7 @@ Content-Length: 57
 
 -----
 ```
-* The boundary used is `-` and is passed to the client in the http response's content-type header. 
+* The boundary used is `-` and is passed to the client in the http response's content-type header. Note that headers can appear in both the HTTP response itself and as part of the response body. The content-type header must be sent in the HTTP response.
```
benjie, line 34:
```suggestion
* The boundary used is `-` and is passed to the client in the http response's `Content-Type` header. Note that headers can appear in both the HTTP response itself and as part of the response body. The `Content-Type` header must be sent in the HTTP response.
```

benjie (approved):
:+1:

PR #160, opened 2021-05-26
Make sure the variables example is valid JSON

phillip-kruger:
The example variables input is not valid JSON. (missing "")

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -156,7 +156,7 @@ query ($id: ID!) {
 
 With the following query variables:
 
-```graphql example
+```json example
 {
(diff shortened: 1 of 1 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -158,7 +158,7 @@ With the following query variables:
 
 ```graphql example
```
benjie, line 159: The language could be improved too:

```suggestion
```json example
```

phillip-kruger: Agree ! Or maybe "variables as json input example" ?

spawnia (approved):

PR #83, opened 2020-04-24
Provide clarifications on content type and serialization

sjparsons:
This PR introduces some new rules in addition to addressing clarifications on the behavior of servers and clients with response to content types. 

Most notably:

1. Specifies that servers MUST support JSON serialization and the associated `application/graphl+json` content type.
2. Describes how servers and clients should interact with respect to content types.
3. Separates examples in both GET and POST requests to their own section heading

PR #161, opened 2021-05-31
Provide clarifications on content type and serialization

ghmcadams:
This PR is an attempt to revive #83 by @sjparsons while modifying as little as possible.

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -112,13 +112,29 @@ http://example.com/product/graphql
 
 # Serialization Format
 
-Similar to the GraphQL specification this specification does not require a specific serialization format.
+The GraphQL specification allows for many [serialization formats to be implemented](https://spec.graphql.org/June2018/#sec-Serialization-Format). Servers and clients over HTTP MUST support JSON and MAY support other, additional serialization formats.
+
 For consistency and ease of notation, examples of the response are given in JSON throughout the spec.
+
+## Content Types
+
+The following are the officially recognized GraphQL content types to designate encoding JSON over HTTP.
+
+| Name | Description |
+| `application/graphql+json` | Required
+| `application/json` | To support legacy clients
+
+A server MUST support requests from clients with HTTP header `Content-Type: application/graphql+json` indicating that the body of the request is a JSON document with a GraphQL request. A server must indicate the content type of the response with a `Content-Type` header. This default value should be `Content-Type: application/graphql+json` indicating that the response is a valid JSON document containing at least one of the `data` or `errors` as a top-level attribute.
+
+A server MAY support requests from clients with other content types.
+
+A client MUST handle receiving responses with HTTP header `Content-Type: application/graphql+json` since any compliant server must support this type.
+
+This header MAY include encoding information (e.g. `Content-Type: application/graphql+json; charset=utf-8`)
 
 # Request
 
-The GraphQL HTTP server SHOULD handle the HTTP GET and POST methods.
-Additionally, GraphQL MAY be used in combination with other HTTP request methods.
+A server MUST accept POST requests, and MAY accept other HTTP methods, such as GET.
 
 ## Request Parameters
 
@@ -129,6 +145,8 @@ A request for execution should contain the following request parameters:
 * {variables} - (*Optional*): Values for any Variables defined by the Operation.
 * {extensions} - (*Optional*): This entry is reserved for implementors to extend the protocol however they see fit.
 
+Note: Be aware that `query` is a misleading name as it can contain a string describing the operation that may be a query, a mutation or a subscription.
+
 Note: depending on the serialization format used, values of the aforementioned parameters can be
 encoded differently but their names and semantics must stay the same.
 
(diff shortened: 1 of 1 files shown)


PowerKiKi (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -112,13 +112,29 @@ http://example.com/product/graphql
...
+
+A client MUST handle receiving responses with HTTP header `Content-Type: application/graphql+json` since any compliant server must support this type.
+
+This header MAY include encoding information (e.g. `Content-Type: application/graphql+json; charset=utf-8`)
+
 # Request
 
-The GraphQL HTTP server SHOULD handle the HTTP GET and POST methods.
-Additionally, GraphQL MAY be used in combination with other HTTP request methods.
+A server MUST accept POST requests, and MAY accept other HTTP methods, such as GET.
```
benjie, line 137: A read-only GraphQL API that only uses persisted queries may use only GET; however by virtue of only supporting persisted queries its not really a full GraphQL server so I think we can keep this proposed change 👍 (otherwise POST is required as GET has size limits that vary wildly)

spec/GraphQLOverHTTP.md
```diff
@@ -129,6 +145,8 @@ A request for execution should contain the following request parameters:
 * {variables} - (*Optional*): Values for any Variables defined by the Operation.
 * {extensions} - (*Optional*): This entry is reserved for implementors to extend the protocol however they see fit.
 
+Note: Be aware that `query` is a misleading name as it can contain a string describing the operation that may be a query, a mutation or a subscription.
```
benjie, line 148:
```suggestion
Note: Be aware that `query` is a misleading name as it can contain a string describing multiple operations, each of which may be a query, mutation or subscription. A better name would have been `document`, but the term `query` is well established.
```

spec/GraphQLOverHTTP.md
```diff
@@ -184,18 +207,26 @@ For example if the `Content-Type` is `application/json` then the request body ma
...
-When a GraphQL server receives a request, it must return a well‐formed response. The server’s
+When a GraphQL server receives a request, it must return a well‐formed response. The server's
 response describes the result of executing the requested operation if successful, and describes
 any errors encountered during the request.
 
 ## Body
 
 If the server's response contains a body it should follow the requirements for [GraphQL response](https://graphql.github.io/graphql-spec/June2018/#sec-Response).
 
+A server MUST return a `Content-Type` HTTP Header with a value of a valid GraphQL content type. By default the response MUST be serialized as JSON and MUST include a  "Content-Type: application/graphql+json` header.
```
benjie, line 220:
```suggestion
A server MUST return a `Content-Type` HTTP Header with a value of a valid GraphQL content type. If there is no `Accept` header in the request, the response MUST be serialized as JSON and MUST include a  "Content-Type: application/graphql+json` header.
```

spec/GraphQLOverHTTP.md
```diff
@@ -184,18 +207,26 @@ For example if the `Content-Type` is `application/json` then the request body ma
...
+
+If another content type is preferable to a client, it MAY include an `Accept` HTTP header listing other acceptable content types in order of preference. In this case a client SHOULD include `application/graphql+json` in the list, according to their preferred priority.
+If no `Accept` header is given, the server MUST respond with a content type of `application/graphql+json`.
```
benjie, line 223:
```suggestion
```

benjie (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -112,13 +112,29 @@ http://example.com/product/graphql
...
+The GraphQL specification allows for many [serialization formats to be implemented](https://spec.graphql.org/June2018/#sec-Serialization-Format). Servers and clients over HTTP MUST support JSON and MAY support other, additional serialization formats.
+
 For consistency and ease of notation, examples of the response are given in JSON throughout the spec.
 
+## Content Types
+
+The following are the officially recognized GraphQL content types to designate encoding JSON over HTTP.
+
+| Name | Description |
+| `application/graphql+json` | Required
```
spawnia, line 124:
```suggestion
| `application/graphql+json` | Required |
```

spec/GraphQLOverHTTP.md
```diff
@@ -112,13 +112,29 @@ http://example.com/product/graphql
...
+| `application/json` | To support legacy clients
```
spawnia, line 125:
```suggestion
| `application/json` | To support legacy clients |
```

spec/GraphQLOverHTTP.md
```diff
@@ -112,13 +112,29 @@ http://example.com/product/graphql
...
+
+A server MUST support requests from clients with HTTP header `Content-Type: application/graphql+json` indicating that the body of the request is a JSON document with a GraphQL request. A server must indicate the content type of the response with a `Content-Type` header. This default value should be `Content-Type: application/graphql+json` indicating that the response is a valid JSON document containing at least one of the `data` or `errors` as a top-level attribute.
```
spawnia, line 127: The last part could refer to the relevant spec instead of reiterating it in a simplified form.

spec/GraphQLOverHTTP.md
spawnia: The standard Introspection query would fit within a commonly used limit of 2048 characters, so that fundamental aspect of GraphQL could work. Still, for interoperability I think supporting POST is important. Otherwise it would complicate the implementation of clients, forcing them to have a fallback mechanism.

spec/GraphQLOverHTTP.md
```diff
@@ -138,13 +156,15 @@ Note: {variables} and {extensions}, if set, must have a map as its value.
 
 ## GET
 
-For HTTP GET requests, the query parameters should be provided in the query component of the request URL in the form of
+For HTTP GET requests, the query parameters MUST be provided in the query component of the request URL in the form of
 `key=value` pairs with `&` symbol as a separator and both the key and value should have their "reserved" characters percent-encoded as specified in [section 2 of RFC3986](https://tools.ietf.org/html/rfc3986#section-2).
-The unencoded value of the {variables} parameter should be represented as a JSON-encoded string.
+The `variables` parameter should be represented as a URL-encoded JSON string.
```
spawnia, line 161:
```suggestion
The `variables` parameter, if used, MUST be represented as a URL-encoded JSON string.
```

What is a server supposed to do with any other encoding, e.g. a URL-encoded XML string?

PR #162, opened 2021-06-01
Fix markdown table

ghmcadams:


benjie:
Recommend you use `prettier` for formatting the markdown; that will autofix tables/etc. Latest releases of `spec-md` now have prettier support.

PR #163, opened 2021-08-21
GraphQL over Server-Sent Events

enisdenjo:
#### Implementations

- [graphql-sse](https://github.com/enisdenjo/graphql-sse)
- _Feel free to add your implementation here!_


benjie:
As this is an RFC it should be merged. (I cannot merge it, but I would if I could.)

PR #166, opened 2021-11-19
remove subscriptions from allowed POST operations

n1ru4l:
When executing a subscription operation a client expects multiple results which can best be expressed as a stream, while the incremental delivery RFC could be misused for describing how a stream of subscription events can be delivered to clients, it is clearly not the correct fit, as multipart requests are intended to be short-lived, while subscriptions could stay active for longer periods of time. Until alternative directions such as SSE events (as supported via graphql-helix and graphql-sse) are explored it might make sense to completely remove `subscriptions` from the list of allowed operations that can be executed via `POST`.

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -194,7 +194,7 @@ GET requests MUST NOT be used for executing mutation operations. If the values o
 
 ## POST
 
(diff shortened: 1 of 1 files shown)


benjie (approved):
Agree. Further many servers when you POST a subscribe operation will `execute` it rather than `subscribe`, and that rarely makes sense to do.

ghmcadams (approved):

PR #170, opened 2022-04-17
Fix POST docs

sungam3r:


ghmcadams (approved):

benjie:
The behaviour of subscription over POST seems under-specified. I'd suggest removing `subscription` for now until that is cleared up. I believe express-graphql does the `execute` call without doing the `subscribe` call if it sees a subscription over POST, but I do not believe this behaviour should be considered correct (and may well have changed since I last checked years ago).

ghmcadams:
Closing this due to the lack of response to comments

PR #174, opened 2022-06-08
Format documents with 'prettier'

benjie:
This copies over some of the standard setup from https://github.com/graphql/wg-template, notably the use of `prettier` to format the documents. This is, in part, because the GraphQL specification itself is now formatted with prettier.

You may want to review the commits individually, the last commit is simply the result of `yarn format`.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+/working-group/
+/rfcs/
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,4 @@
+# PRs to Working Group documents should be minimal effort - no need to force formatting compliance
 /working-group/
+# To maintain the expressiveness of the original poster, we do not auto-format RFCs
 /rfcs/
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,15 @@ typings/
 
 # Visual Studio project specific, machine local files
 .vs/
+
+# Standard ignores from wg-template
+*.swp
+*~
+.*.haste_cache.*
+.DS_Store
+npm-debug.log
+/build
+/public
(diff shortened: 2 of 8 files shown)


.prettierignore
```diff
@@ -0,0 +1,2 @@
+/working-group/
+/rfcs/
```
spawnia, line 2: Why shouldn't those be formatted?

benjie: Generally I find it's easier if the things that people like to edit through GitHub are not formatted via prettier (because CI failing on them is annoying) but the main spec text itself should be nicely formatted. That, and the size of the changes was _massive_ for these folders!

benjie: I'm definitely open to formatting everything with prettier though; we can slap a `<!-- prettier-ignore -->` in front of the WG attendees lists so it doesn't become a pain.

ghmcadams: I believe RFCs should not be formatted.  They are written as a way of communicating and should not be altered by anyone other than the original poster.

ghmcadams (approved):

.prettierignore
spawnia: I agree, perhaps a short code comment would be good to document that thinking.

benjie, line 2:
```suggestion
# PRs to Working Group documents should be minimal effort - no need to force formatting compliance
/working-group/
# To maintain the expressiveness of the original poster, we do not auto-format RFCs
/rfcs/
```

benjie: Comments added

benjie:
Since this gained approval and is adding noise to #175 I'm going to go ahead and merge it :+1:

PR #175, opened 2022-06-08
Rework specification with a focus on backwards compatibility

benjie:
**TL;DR: Rather than reading this PR as a diff, please just read the finished content. Sorry :heart:**

First, I'm sorry. I had to back out of involvement in the GraphQL-over-HTTP working group a couple years ago due to severe lack of time and I'm only just getting my head back above water now. I'm keen for us to start making progress again, but it seems we have quite a number of sticking points, not least the contention over `application/graphql+json` (which, if I recall correctly, was my suggestion in the first place).

Originally this specification was intended to codify existing practices such that the majority of existing servers and clients were already compatible. It turned out that HTTP status codes were a big sticking point, and this lead us to introduce the `application/graphql+json` media type. We marked this media type as required from the get-go.

In this revised proposal, I instead introduce a "watershed" of 1st Jan 2025 whereupon support from `application/graphql+json` changes from "RECOMMENDED" to "REQUIRED".

I have also tightened up the language, referenced the relevant RFCs and spec locations, and in many places _concretely changed the behavior that was previously specified_. I've also restructured the specification significantly.

This is not a simple editorial pull request, nor one made up of incremental changes that can be easily separated into separate smaller PRs; this is pretty much a re-write of the specification, and for that I am very sorry. I have spent a decent number of hours going through everything with a fine-toothed comb, but this is still just a first draft and I expect there to be changes in follow-up PRs even after (if) this is merged.


spec/GraphQLOverHTTP.md
```diff
@@ -70,33 +90,70 @@ Note: This is an example of a non-normative note.
...
 
-Serving GraphQL over HTTP provides the ability to use the full advantages of GraphQL with the rich feature set of HTTP.
-Carrying GraphQL in HTTP does not mean that GraphQL overrides existing semantics of HTTP,
-but rather that the semantics of GraphQL over HTTP map naturally to HTTP semantics.
+Though [the GraphQL specification](https://spec.graphql.org) is transport
+agnostic, this GraphQL over HTTP specification aims to map GraphQL's semantics
+to their HTTP equivalents, enabling us to combine the full advantages of GraphQL
+with the rich feature set of HTTP.
+
+GraphQL queries and mutations naturally follow the request/response message
```
benjie, line 98:
```suggestion
GraphQL queries and mutations naturally mirror the request/response message
```

spec/GraphQLOverHTTP.md
```diff
@@ -187,94 +305,298 @@ This request could be sent via an HTTP GET as follows:
...
+`application/json` media type.
 
-### Unparseable or invalid request body
+The server SHOULD use the `200` status code, independent of any {GraphQL request
+error} or {GraphQL field error} raised.
+
+Note: A status code in the `4xx` or `5xx` ranges or status code `203` (and maybe
+others) could originate from intermediary servers; since the client cannot
+determine if an `application/json` response with arbitrary status code is a
+well-formed GraphQL response (because it cannot trust the source) we must use
```
ghmcadams, line 482: Should `we` be replaced with something else?

spec/GraphQLOverHTTP.md
```diff
@@ -70,33 +90,70 @@ Note: This is an example of a non-normative note.
...
-Clients can discover the schema by sending an introspection query, which the server MUST answer with a proper response. 
+A client MUST be able to send all their GraphQL query and mutation requests for
+a single GraphQL schema made available by a server to a single URL endpoint on
+that server. A server MUST NOT require the client to use different URLs for
+different GraphQL query and mutation requests to the same GraphQL schema.
 
-The schema on a single URL MAY not be the same for every client.
-For example, certain fields could be restricted to authenticated users or alpha testers.
+Note: This means that a client must be able to issue all GraphQL query and
+mutation operations it needs via a single endpoint.
```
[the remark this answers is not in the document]
benjie: Good question. The first two are required; imagine you have three endpoints for the same schema: `/a`, `/b` and `/c`.

Imagine you can issue all operations to `/a`, only named operations to `/b` and only unnamed operations to `/c`. This would pass sentence 1 but not sentence 2. But sentence 2 might cover both cases.

I think you're right; we should trim this down to just sentence 2.

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +169,133 @@ http://example.com/product/graphql
...
-| ------------- | ------------- |
-| `application/graphql+json` | Required |
-| `application/json` | To support legacy clients |
+The following are the officially recognized GraphQL media types to designate
+using the JSON encoding for GraphQL requests and responses over HTTP.
 
-A server MUST support requests from clients with HTTP header `Content-Type: application/graphql+json` indicating that the body of the request is a JSON document with a GraphQL request. A server must indicate the content type of the response with a `Content-Type` header. This default value should be `Content-Type: application/graphql+json` indicating that the response is a valid JSON document [conforming to the GraphQL spec](https://spec.graphql.org/October2021/#sec-Response-Format).
+| Name                       | Description                                   |
+| -------------------------- | --------------------------------------------- |
+| `application/graphql+json` | The preferred type; better HTTP compatibility |
```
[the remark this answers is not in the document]
benjie: I think that's outside the scope of this spec right now; we do _allow_ for it, but we don't _specify_ it.

spec/GraphQLOverHTTP.md
```diff
@@ -70,33 +90,70 @@ Note: This is an example of a non-normative note.
...
+A server MUST enable GraphQL requests to one of more GraphQL schemas.
+
+Each GraphQL schema a server provides MUST be served via one or more URLs.
 
-A GraphQL schema allows clients to know the available operations on a GraphQL server.
-Clients can discover the schema by sending an introspection query, which the server MUST answer with a proper response. 
+A client MUST be able to send all their GraphQL query and mutation requests for
+a single GraphQL schema made available by a server to a single URL endpoint on
+that server. A server MUST NOT require the client to use different URLs for
+different GraphQL query and mutation requests to the same GraphQL schema.
```
benjie, line 132:
```suggestion
A server MUST NOT require the client to use different URLs for
different GraphQL query and mutation requests to the same GraphQL schema.
```

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +169,133 @@ http://example.com/product/graphql
...
 
-Note: specifying `null` in JSON (or equivalent values in other formats) as values for optional request parameters is equivalent to not specifying them at all.
+Note: Specifying `null` in JSON (or equivalent values in other formats) as
+values for optional request parameters is equivalent to not specifying them at
+all.
 
-Note: {variables} and {extensions}, if set, must have a map as its value.
+Note: {variables} and {extensions}, if set, must have a map as their value.
+
+### JSON Encoding
```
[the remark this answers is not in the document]
benjie: Maybe, but I moved it up here because it's one of a number of possible encodings of the request parameters; for `GET` there's only really one choice. But yeah, maybe we should move it back again :man_shrugging:

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
ghmcadams: I think we should state something, though.  A server that chooses not to support JSON would not know what to do.

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +169,133 @@ http://example.com/product/graphql
...
+
+Note: This recommended header enables compatibility with legacy servers whilst
+still leveraging modern features if available in the server.
 
 ## GET
 
-For HTTP GET requests, the query parameters MUST be provided in the query component of the request URL in the form of
-`key=value` pairs with `&` symbol as a separator and both the key and value should have their "reserved" characters percent-encoded as specified in [section 2 of RFC3986](https://tools.ietf.org/html/rfc3986#section-2).
-The `variables` parameter, if used, MUST be represented as a URL-encoded JSON string.
+For HTTP GET requests, the query parameters MUST be provided in the query
```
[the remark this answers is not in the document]
benjie: Good catch! I'm surprised the word "query" didn't stick out to me more :rofl:

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +169,133 @@ http://example.com/product/graphql
...
+
+- {query} - the string representation the Source Text of the Document as
+  specified in
+  [the Language section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Language).
+- {operationName} - an optional string
+- {variables} - an optional object (map), the keys of which are the variable
+  names and the values of which are the variable values
+- {extensions} - an optional object (map)
+
+## Accept
```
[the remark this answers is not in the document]
ghmcadams: This spec bounces around a lot.  If I were implementing a client or a server separately, it might be hard to determine what it means to be compliant.  Creating a separation might make it easier.

spec/GraphQLOverHTTP.md
```diff
@@ -187,94 +305,298 @@ This request could be sent via an HTTP GET as follows:
...
+Allowed) and halt execution. This restriction is necessary to conform with the
+long-established semantics of safe methods within HTTP.
 
 ## POST
 
-A GraphQL POST request instructs the server to perform a query or mutation operation. A GraphQL POST request MUST have a body which contains values of the request parameters encoded according to the value of `Content-Type` header of the request.
+A GraphQL POST request instructs the server to perform a query or mutation
+operation. A GraphQL POST request MUST have a body which contains values of the
+request parameters encoded in one of the officially recognized GraphQL media
+types, or another media type supported by the server.
```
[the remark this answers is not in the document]
benjie: I see your point. I think that though JSON is the concrete representation, because there are two distinct media types (`application/json` and `application/graphql+json`) it makes sense to keep it as is - but if you can think of better wording I'm definitely open to it!

spec/GraphQLOverHTTP.md
```diff
@@ -187,94 +305,298 @@ This request could be sent via an HTTP GET as follows:
...
-any errors encountered during the request.
+When a server receives a GraphQL request, it must return a well‐formed response.
+The server's response describes the result of validating and executing the
+requested operation if successful, and describes any errors encountered during
+the request.
+
+A server must comply with
+[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
+
+## Interpreting the request
```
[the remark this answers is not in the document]
benjie: Yeah; you might be right. I moved this around quite a few times before putting it here; I think I was thinking of it as "client" and "server" rather than "request" and "response" but the sections are clearly request/response so I should move it.

spec/GraphQLOverHTTP.md
```diff
@@ -187,94 +305,298 @@ This request could be sent via an HTTP GET as follows:
...
+A server SHOULD support requests which accept the `application/graphql+json`
+media type (as indicated by the `Accept` header).
+
+A server MAY support requests encoded with and/or accepting other media types.
+
+If the client does not supply a `Content-Type` header, the server SHOULD treat
+the request as if it had `Content-Type: application/json`.
+
+If the client does not supply an `Accept` header, the server SHOULD treat the
+request as if it had `Accept: application/json`.
```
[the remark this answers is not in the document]
benjie: Right, but the `Note` is non-normative - it's mostly there for people just reading that part of the spec and not finding this. We may not need it if we move this section.

spec/GraphQLOverHTTP.md
```diff
@@ -187,94 +305,298 @@ This request could be sent via an HTTP GET as follows:
...
 
-1. Disregard the `Accept` header and respond with the default content type of `application/graphql+json`, specifying this in the `Content-Type` header; OR
-2. Respond with a `406 Not Acceptable` status code
+## Execution
 
-Note: For any non-`2XX` response, the client should not rely on the body to be in GraphQL format since the source of the response
-may not be the GraphQL server but instead some intermediary such as API gateways, firewalls, etc.
+Execution of a GraphQL-over-HTTP request follows the
+[ExecuteRequest()](<https://spec.graphql.org/draft/#ExecuteRequest()>) algorithm
+specified in the GraphQL specification.
```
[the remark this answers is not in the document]
benjie: I'm not sure exactly how to do that, and doing so may confuse terminology (e.g. "request" would become heavily overloaded). I think better to keep them separate and reference.

I've used "GraphQL's ExecuteRequest()" style of wording in other places that seems to work well (and be less wordy) - I'll adopt that here :+1:

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
benjie: I'm going to move "A server MAY support requests encoded with and/or accepting other media types." and related rules back into this section.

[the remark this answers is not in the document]
benjie: (I've made this change; if you're happy with it please close the conversation :+1:)

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
benjie: Moved.

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
benjie: Moved into the "Media type" section instead.

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
benjie: Ah, great comment. Believe it or not I was trying not to change too much so didn't change the section titles :laughing: But yeah, I think I was thinking of it in terms of client/server rather than request/response a lot of the time, it feels more natural. I'll think about it...

benjie:
Thanks for the careful read @ghmcadams! I've applied a number of your suggestions and closed those discussions; others are left open to gather more feedback :+1:

spec/GraphQLOverHTTP.md
```diff
@@ -122,17 +122,12 @@ Specification compliant HTTP client unless the context indicates otherwise.
...
 
-A client MUST be able to send all their GraphQL query and mutation requests for
-a single GraphQL schema made available by a server to a single URL endpoint on
-that server. A server MUST NOT require the client to use different URLs for
-different GraphQL query and mutation requests to the same GraphQL schema.
-
-Note: This means that a client must be able to issue all GraphQL query and
-mutation operations it needs via a single endpoint.
+A server MUST NOT require the client to use different URLs for different GraphQL
+query and mutation requests to the same GraphQL schema.
```
ghmcadams, line 130: By mentioning these specifically, it might create confusion as to why subscriptions are not mentioned.

```suggestion
A server MUST NOT require the client to use different URLs for different GraphQL
requests to the same GraphQL schema.
```

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
ghmcadams: I see your point.  And this style of wording does work well.  It got me to thinking, though, that there is a big assumption for this spec that has been left unsaid (that this spec seeks never to override the GraphQL spec, but only add to it in order to support serving it over HTTP).

Readers may (if left unsaid) be confused at times if ever there is something in this spec that seems to be in conflict with the GraphQL spec.  For instance, there is no distinction in this spec between an HTTP server and a GraphQL server (perhaps an implementation would combine them, but perhaps not).  If an implementor were to read this spec, and see that it states that `The GraphQL server should not make authorization decisions based on any part of the GraphQL request`, they may be confused.

ghmcadams:
> does the concept of "server" encompass the GraphQL part or not?  My intent is that it "wraps" the GraphQL part and almost sees GraphQL as a "black box" (except as noted by specific algorithm references), but finding a way of stating this does seem desirable. Let me know if you have specific thoughts how to address this.

I agree.  To address this, this specification could include these additional statements:

### Introduction
`Replace the first paragraph with this:`

This specification details how libraries that wish to publish and consume GraphQL services over HTTP should do so in order to maximize interoperability between client libraries, tools, and server implementations.  The goal of this specification is not to override or replace the [GraphQL specification](https://spec.graphql.org/).  It seeks only to augment it for the purpose of serving GraphQL services over HTTP.  If ever there is conflict between the two, the GraphQL specification should be used.


### Overview

:: In this document, the term {GraphQL service} refers to an [application service as used by the GraphQL specification](https://spec.graphql.org/draft/#sec-Overview).

`just after the definitions of {server} and {client} respectively:`

The role of a server is to provide clients access to one or more GraphQL services over HTTP.  A server is not a GraphQL service.  It is a GraphQL service host.

`Optional (though I don't foresee there being ambiguities in this area):`

The role of a client is to issue requests to a server over HTTP in order to interact with a GraphQL service.

benjie:
Excellent suggestions @ghmcadams; I've incorporated wording similar to that into the document :+1:

benjie:
The topic of request/response versus client/server is one I would like to discuss in a working group. Let's organise one in the next month or two.

ghmcadams (approved):

ghmcadams:
Great work @benjie.  This spec looks awesome.

ROADMAP.md
```diff
@@ -43,42 +55,52 @@ Future versions of the spec may include these concepts:
...
-The process of writing this specification may proceed according this rough outline of stages.
-We are currently in the *Preliminary Stage*.
+The process of writing this specification may proceed according this rough
+outline of stages. We are currently in the _Preliminary Stage_.
 
 ### Stage 0: Preliminary
 
-In the *Preliminary Stage*, things may change rapidly and no one should count on any particular
-detail remaining the same.
+In the _Preliminary Stage_, things may change rapidly and no one should count on
```
michaelstaib, line 75: should we align the names of the stages with the main spec?

ROADMAP.md
```diff
@@ -43,42 +55,52 @@ Future versions of the spec may include these concepts:
...
+any particular detail remaining the same.
 
-- If a PR has no requests for changes for 2 weeks then it should be merged by one of the maintainers
-- If anyone has an objection later, they just open a PR to make the change and it goes through the same process
+- If a PR has no requests for changes for 2 weeks then it should be merged by
```
michaelstaib, line 78: Is the thinking here to have some pressure for people to at least formulate some objections?

ghmcadams: The goal was to keep the process moving.  At this stage, we did not want a PR to be sitting for too long.

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +178,152 @@ http://example.com/product/graphql
...
-A server MUST accept POST requests, and MAY accept other HTTP methods, such as GET.
+A server MUST accept POST requests, and MAY accept other HTTP methods, such as
+GET.
 
 ## Request Parameters
 
-A request for execution should contain the following request parameters:
+A {GraphQL-over-HTTP request} is formed of the following parameters:
+
+- {query} - A Document containing GraphQL Operations and Fragments to execute.
```
michaelstaib, line 257: This also should be optional to allow for persisted queries, but I can do a separate PR on this outlining the use-case. I original added the `{extensions}` for this so the persisted query id can now go into that map.

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +178,152 @@ http://example.com/product/graphql
...
+- {variables} - (_Optional_): Values for any Variables defined by the Operation.
+- {extensions} - (_Optional_): This entry is reserved for implementors to extend
+  the protocol however they see fit.
+
+Note: When comparing against
+[request in the GraphQL specification](https://spec.graphql.org/draft/#request)
+you should note the schema and "initial value" are not included in the
+GraphQL-over-HTTP request, they are handled by the server based on the URL used.
+
+Note: Be aware that `query` is a misleading parameter name as its value is a
```
michaelstaib, line 269: Should we not introduce a new term in this case `document` and support `query` as deprecated. Also, this discussion should go in a separate PR. But something we should think about.

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +178,152 @@ http://example.com/product/graphql
...
 
-For consistency and ease of notation, examples of the response are given in JSON throughout the spec.
+A server SHOULD support requests which accept the `application/graphql+json`
+media type (as indicated by the `Accept` header).
 
-## Content Types
+A server MAY support requests encoded with and/or accepting other media types.
 
-The following are the officially recognized GraphQL content types to designate encoding JSON over HTTP.
+If the client does not supply a `Content-Type` header, the server SHOULD treat
```
michaelstaib, line 221: Do we have servers out there that accept the `Content-Type` being empty? I think the biggest issue was more about the Accept header since many clients do not yet send one. 

I think we can expect the `Content-Type` header to be set to `application/json` or `application/graphql+json`. Legacy clients will send `application/json` and this SHOULD be supported.

spec/GraphQLOverHTTP.md
benjie: I don't think that a server that only allows persisted operations is a truly GraphQL-over-HTTP compliant server. The literal intent of an operation allowlist is to _prevent_ interoperability, and that's okay!

spec/GraphQLOverHTTP.md
benjie: I was seriously tempted to make it so that `application/graphql+json` uses the term "document" (or maybe "source") whereas `application/json` uses "query", but decided that this would probably be ill advised. So yeah, definitely a follow-up discussion for that 👍

spec/GraphQLOverHTTP.md
benjie: You're probably right; even if you're using `curl` you'd expect to set a content-type header, otherwise it's assumed it's `application/x-www-form-urlencoded`. Good catch; I'll get this fixed 👍

spec/GraphQLOverHTTP.md
```diff
@@ -1,57 +1,82 @@
...
-The closest thing to an official specification is the article [Serving over HTTP](https://graphql.org/learn/serving-over-http/).
-Leading implementations on both client and server have mostly upheld those best practices and thus established
-a de-facto standard that is commonly used throughout the ecosystem.
-
-This specification is intended to fill this gap by specifying how GraphQL should be served over HTTP.
-The main intention of this specification is to provide interoperability between different client libraries, tools
-and server implementations.
+This specification details how libraries that wish to publish and consume
+GraphQL services over HTTP should do so in order to maximize interoperability
+between client libraries, tools, and server implementations. This specification
```
spawnia, line 19:
```suggestion
This specification details how GraphQL should be served and consumed over HTTP
in order to maximize interoperability between clients, servers and tools. This specification
```

The spec concerns not only libraries, but also frameworks, applications, etc.

spec/GraphQLOverHTTP.md
```diff
@@ -1,57 +1,82 @@
...
+both client and server have mostly upheld those best practices and thus
+established a de-facto standard that is commonly used throughout the ecosystem.
+This specification aims to codify and expand on this work.
 
 **Copyright notice**
 
-**TBD**
+Copyright © 2022-present, GraphQL contributors
+
+THESE MATERIALS ARE PROVIDED “AS IS.” The parties expressly disclaim any
```
spawnia, line 42:
```suggestion
THESE MATERIALS ARE PROVIDED “AS IS”. The parties expressly disclaim any
```

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +178,152 @@ http://example.com/product/graphql
...
+| `application/graphql+json` | The preferred type; better HTTP compatibility |
+| `application/json`         | To support legacy clients                     |
+
+For details of the shapes of these JSON payloads, please see
+[Request](#sec-Request) and [Response](#sec-Response).
+
+If the media type in a `Content-Type` or `Accept` header includes encoding
+information, then the encoding MUST be `utf-8` (e.g.
+`Content-Type: application/graphql+json; charset=utf-8`). If encoding
+information is not included then `utf-8` should be assumed.
```
spawnia, line 205:
```suggestion
information is not included then `utf-8` MUST be assumed.
```

I think making this default a strong requirement is essential to ensure interoperability.

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +178,152 @@ http://example.com/product/graphql
...
+A {GraphQL-over-HTTP request} is formed of the following parameters:
+
+- {query} - A Document containing GraphQL Operations and Fragments to execute.
+- {operationName} - (_Optional_): The name of the Operation in the Document to
+  execute.
+- {variables} - (_Optional_): Values for any Variables defined by the Operation.
+- {extensions} - (_Optional_): This entry is reserved for implementors to extend
+  the protocol however they see fit.
+
+Note: When comparing against
```
spawnia, line 264:
```suggestion
Note: When comparing against a
```

spec/GraphQLOverHTTP.md
spawnia, line 257:
```suggestion
- {query} - The string representation of a Document containing GraphQL Operations and Fragments to execute.
```

See https://github.com/graphql/graphql-over-http/issues/176 for context.

spec/GraphQLOverHTTP.md
```diff
@@ -112,72 +178,152 @@ http://example.com/product/graphql
...
-    name
-  }
-}
+```raw graphql example
+query($id: ID!){user(id:$id){name}}
 ```
 
 With the following query variables:
 
 ```json example
```
spawnia, line 324: Should this also be in the raw encoded form?

spec/GraphQLOverHTTP.md
```diff
@@ -187,94 +333,277 @@ This request could be sent via an HTTP GET as follows:
...
+Note: If the client does not include a `Content-Type` header then
+`application/json` will be assumed.
 
-A client MUST include a `Content-Type` with a POST request.
+### JSON Encoding
+
+When encoded in JSON, a GraphQL-over-HTTP request is a JSON object (map), with
+the properties specified by the GraphQL-over-HTTP request:
+
+- {query} - the string representation the Source Text of the Document as
```
spawnia, line 367:
```suggestion
- {query} - the string representation of the Source Text of the Document as
```

spec/GraphQLOverHTTP.md
```diff
@@ -187,94 +333,277 @@ This request could be sent via an HTTP GET as follows:
...
+```headers example
+Content-Type: application/graphql+json
+Accept: application/graphql+json
+```
+
+And the body:
+
+```json example
+{
+  "query": "query ($id: ID!) { user(id:$id) { name } }",
```
spawnia, line 415:
```suggestion
  "query": "query($id:ID!){user(id:$id){name}}",
```

Or add in `\n` and whitespace to make it equivalent to the previous example

spec/GraphQLOverHTTP.md
benjie: What would be a good catch-all term? `software`?

ghmcadams: `implementors`?

I also like the suggested text.

spec/GraphQLOverHTTP.md
spawnia: Not necessarily. I checked the MIT license - which I believe inspired this - and it puts a comma behind the quotes. https://opensource.org/licenses/MIT

spec/GraphQLOverHTTP.md
spawnia: If we phrase it different, we can avoid having to find a term.

spec/GraphQLOverHTTP.md
ghmcadams: Agreed

spec/GraphQLOverHTTP.md
ghmcadams: Though it does sound like it should read `When comparing against a request...`, I don't think it should.  I believe it should be read as something like `When comparing against the request section in the GraphQL specification`.

glasser:
This is exciting! Going to try to find time this week to look at it from an Apollo Server perspective.

michaelstaib:
This already looks great @benjie ... will give it a third read tomorrow :)

benjie:
I have rebased this PR on top of #174; I did try a simple merge but there were too many conflicts.

spec/GraphQLOverHTTP.md
benjie: My apologies I completely missed that there was a suggestion there somehow :sweat_smile:

spec/GraphQLOverHTTP.md
benjie: This copyright notice is copied and pasted from the GraphQL spec; if we are to change it, we should change it there before reflecting it elsewhere:

https://github.com/graphql/graphql-spec/blob/4de87822d2bc807b746c6a37f6a99bb54d705185/spec/GraphQL.md?plain=1#L35

But although the grammar is incorrect, I agree that we should use the standard license, not the one with correct grammar :+1: 

I'm going to close this conversation, but please feel free to raise a PR against the GraphQL spec and we can discuss it there.

spec/GraphQLOverHTTP.md
benjie: I've changed it such that the `Content-Type` **MUST** be indicated by the client.

spec/GraphQLOverHTTP.md
benjie: I am choosing not to adopt this change for the reasons laid out in https://github.com/graphql/graphql-over-http/issues/176#issuecomment-1154019962 ; let's discuss it in a separate PR.

spec/GraphQLOverHTTP.md
benjie: Indeed, it is referring to the term "request" rather than a specific request. I'll put request in quote marks to make this clearer :+1:

spec/GraphQLOverHTTP.md
spawnia: How is interoperability possible if the client does not send a string?

spec/GraphQLOverHTTP.md
benjie: I simplified the query first, but then decided it was valuable to show `\n` in the encoded data, so I have gone with the whitespace option :+1: Thanks @spawnia

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
benjie: I believe this has now been addressed :+1:

spec/GraphQLOverHTTP.md
spawnia: I agree that inclusion of whitespace makes it obvious that the string can be included as-is and does not need to be compressed. Should we recommend or mention compression in a non-normative section?

spec/GraphQLOverHTTP.md
benjie: In GET, we state:

> The `query` parameter MUST be the string representation of the Source Text of
the Document as specified in [...]

In POST, we state:

> When encoded in JSON [...]
> - {query} - the string representation of the Source Text of the Document as specified in [...]

Thus for the standard encodings, string is specified for compatibility. However, alternative encodings are allowed by the spec (but not specified); these could come up with alternative serialization for the GraphQL query that is more efficient on the wire; for example you could consider a binary encoding that numbers all the keywords known to GraphQL and has an incredibly efficient encoding of a query; this should not fail the GraphQL-over-HTTP spec, it's something that we could later standardize even - e.g. `application/octet-graphql` may one day be included in this specification.

What matters for alternative (non-standard) encodings is what meaning the payloads must contain, not the specific stucture with which they do so.

spec/GraphQLOverHTTP.md
benjie: No; the whitespace in the JSON is unimportant.

benjie:
A decent number of additional edits thanks to @spawnia and @michaelstaib, the most critical of which are:

- Encoding: SHOULD -> MUST assume UTF-8 if unspecified
- Client MUST indicate Content-Type; server SHOULD reject if no Content-Type
- Clarifications around the server not making authorization decisions based on the GraphQL request

benjie:
I've spent the last few minutes restructuring this spec to be in terms of client and server, but there are many shared responsibilities in terms of request and response so I think this would be a larger piece of work. It may be worthwhile, but I suggest that we attempt to land these current changes first before restructuring further. I did at least find a bug whilst doing it, so I've patched that :+1:

spec/GraphQLOverHTTP.md
spawnia: Right, but that is also true for whitespace in the GraphQL. I think it would be self-consistent if both examples were either minified or formatted.

benjie: You're right; the encoding in the URL should match that in the example :+1:

ghmcadams:
@benjie I suggest merging this.  Any further requests can be made by submitting a new PR.

benjie:
@spawnia @michaelstaib How do you feel about merging this?

Are there others still active from the old HTTP WG?

ghmcadams:
@benjie As for your question of active members:  Activity has been low for quite a while.  I think it's safe to merge.  Especially given that the spec is still in [stage 0 (preliminary)](https://github.com/graphql/graphql-over-http/blob/main/ROADMAP.md#stage-0-preliminary), which states that things can change rapidly.  Any objections can be addressed with a new PR.

benjie:
Unless we get a few more approvals, let's give it two weeks and then merge on or after 28th June in line with the "preliminary" policy :+1:

glasser:
I have not yet had time to do a full review of this change yet from an Apollo Server perspective. (This month is very busy for me personally and work-wise and June 28th will be a challenge.)

I'd like to specifically focus on the MIME type issue, though.

My position is that I am in favor of the proposal with respect to response content-types, but I don't find the changes for request content-types to be compelling, for several reasons.

First, I just fundamentally am skeptical of the concept of a single MIME type that means "*either* a JSON object with `query`/`operationName`/`variables` *or* a JSON object with `data`/`errors`/`extensions`". This doesn't feel like a single coherent concept to me.  If you're using a system that lets you write "serialize and deserialize a given MIME type into a given in-memory representation", you'll basically always (in a strongly typed system) need to carefully ensure that your registered handlers only apply to requests or responses, which systems might not support.

Second, my understanding is that the issue we're trying to solve with the new MIME types is to allow clients to tell if a response actually came from a GraphQL server or from some sort of intermediate processor.  That makes sense as a goal to me, and I see how a specific response content-type helps here. But I don't see why having a specific request content-type is at all relevant to the issue we're trying to solve.  For responses, the benefits seem to outweigh the costs to me, but the benefits seem way lower for requests.

Third, for responses there's a great opt-in mechanism via `accept`, which your proposal uses.  Clients who want to solve the main issue (differentiate real GraphQL server responses from intermediate processor responses) can opt in to this by specifying `accept: application/graphql+json`. In many systems (though fortunately not Fetch's `response.json()`, this will also require you to configure to client to understand that `application/graphql+json` should be parsed as JSON, but this is an easy extra hoop to jump through if you care about this feature.  If I understand your proposal correctly, this is *always* going to be an opt-in feature; even after 2025 the default response content-type will still be `application/json`. If a server doesn't implement this feature, the only people who will notice are the folks who are specifically trying to use the feature, and they will be able to go request that their server be upgraded to support the feature; other people will not be affected. There is no corresponding opt-in mechanism for request content-types.

Fourth, requiring support for this new content-type in requests makes implementing a spec-compliant server more complex. Specifically, many GraphQL servers are implemented as middleware that can be layered into a bigger web framework stack. Web frameworks typically have a standard mechanism for parsing incoming requests as JSON which properly solves issues like setting a configurable maximum size to parse, using a standard error handling mechanism, etc.  It is not unreasonable to want to build your GraphQL server as a middleware that assumes the incoming body has already been parsed. In Apollo Server we've historically hardcoded the use of the npm `body-parser` library into our middleware but this has been a bad experience overall and we're dropping that in Apollo Server 4. As another example, [graphql-helix](https://www.npmjs.com/package/graphql-helix) already is implemented as a middleware that expects you to have already run Express's JSON parser. If being a spec-compliant server means understanding multiple request content-types, then a server designed as "middleware that runs after a standard JSON parser step" needs its users to all understand how to configure the middleware to handle both JSON content-types. This step is out of the control of the GraphQL server middleware itself.  Yes, you can document something like `app.use(json({type: ApolloServer.requestContentTypes}))`, but it's easy for a user to forget to do that.  It would be unfortunate to write a spec that makes it easy for end-user error to lead to post-2025 spec violation, just because the GraphQL server is designed in a modular way that takes advantage of the features of the web framework it's embedded in instead of reimplementing those features themselves. (This concern does not apply to response content-types, as it's a lot more typical for middleware to handle JSON output themselves, as there isn't concern about avoiding DOSes via size limits.)

To me, the fourth issue is a pretty big cost to the idea of adding a new request content-type to GraphQL-over-HTTP, and given the lack of a clear problem solved by this request content-type, makes me very unexcited about this aspect of the proposal.

I would advise a simpler approach where we define `application/graphql+json` to mean "a JSON object whose keys may include `data`, `errors`, and `extensions`"; clients may request that responses be sent with this content-type via `accept` header, and by 2025 all servers MUST obey that header if provided. Requests remain `application/json`; no additional burden is placed on end users to configure their web framework's JSON parser properly.

benjie:
Solid argument; I accept.

Whilst making the changes, I also spotted some other issues, for example we said that `Content-Type` was required on requests which doesn't make sense for GET requests. So I added `on POST requests` and then realised they should be in the `POST` section... so I've restructured the spec again :sweat_smile:  (Sorry!)

Changes are, basically:

- Two different tables for the media types; one for requests, one for responses
- Rules on request `Content-Type` moved to Request.POST section
- Rules on request `Accept` moved to Request.Accept section
- Rules on server support for `Accept` header moved to Response.Body section
- Only specify `application/json` for POST request `Content-Type` (`application/graphql+json` no longer explicitly supported as request Content-Type, but of course servers MAY support "other encodings")
- After the watershed, the server _SHOULD_ assume `Accept: application/graphql+json` in the absence of an `Accept` header - this allows status codes to work out of the box. (`Accept: application/json` is still supported if indicated explicitly, resulting in the always-200 behaviour.)
- After the watershed, the requirement that the client SHOULD include `application/json` in `Accept` is lifted.
- Removed the assertion from the "errors should give 4xx" rule into a note (since it's asserted by the GraphQL spec, we do not need to reassert).

benjie:
Assuming no more changes (other than typos), merge date pushed back to 1st July.

spec/GraphQLOverHTTP.md
```diff
@@ -2,38 +2,53 @@
...
+both client and server have mostly upheld those best practices and thus
+established a de-facto standard that is commonly used throughout the ecosystem.
+This specification aims to codify and expand on this work.
 
 **Copyright notice**
 
-**TBD**
+Copyright © 2022-present, GraphQL contributors
+
+THESE MATERIALS ARE PROVIDED “AS IS.” The parties expressly disclaim any
```
spawnia, line 41:
```suggestion
THESE MATERIALS ARE PROVIDED “AS IS”. The parties expressly disclaim any
```

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
...
+types, or another media type supported by the server.
+
+A client MUST indicate the media type of a request body using the `Content-Type`
+header as specified in [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
+
+A server MUST support POST requests encoded with the `application/json` media
+type (as indicated by the `Content-Type` header).
+
+If the client does not supply a `Content-Type` header with a POST request, the
+server SHOULD reject the request using the appropriate `4xx` status code.
```
glasser, line 332: I'd consider making this a MUST for CSRF-protection reasons. A server that allows a POST with no custom headers containing a mutation (and which uses cookies for authentication) is vulnerable to trivial CSRFs.

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
...
+
+A server MAY support POST requests encoded with and/or accepting other media
+types.
```
glasser, line 335: A non-normative note suggesting that servers which support POST requests with content-type `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain` need to take extra care to avoid CSRF may be useful.

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
 
 GET requests MUST NOT be used for executing mutation operations. If the values
 of {query} and {operationName} indicate that a mutation operation is to be
-executed, the server SHOULD immediately respond with error status code `405`
-(Method Not Allowed) and halt execution. This restriction is necessary to
-conform with the long-established semantics of safe methods within HTTP.
+executed, the server MUST respond with error status code `405` (Method Not
+Allowed) and halt execution. This restriction is necessary to conform with the
+long-established semantics of safe methods within HTTP.
```
glasser, line 316: Would it be interesting to include a further non-normative note here saying something like:

Additionally, restricting mutations to POSTs which MUST contain a `Content-Type` header ensures that mutations cannot be sent as part of [cross-site request forgery (CSRF) attacks](https://owasp.org/www-community/attacks/csrf), as long as the `Content-Type` header is not allowed to be one of the three values which can be in a [CORS-safelisted `Content-Type` request header](https://fetch.spec.whatwg.org/#cors-safelisted-request-header) (`application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain`).

(that `MUST` assumes my other suggestion is taken, which perhaps it won't be)

glasser (commented):
If we're interested in making the spec help implementers avoid CSRF vulnerabilities, here are a few thoughts.

benjie:
Good notes @glasser; I think we should add these in a follow-up security focussed PR 👍

spec/GraphQLOverHTTP.md
```diff
@@ -289,11 +254,14 @@ A client SHOULD indicate the media types that it supports in responses using the
 `Accept` HTTP header as specified in
 [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
 
-A client SHOULD include the media type `application/graphql+json` in the
-`Accept` header.
+If the client supplies an `Accept` header, a client SHOULD include the media
```
ghmcadams, line 257: Consider consistent language:  `the client` vs `a client`

spec/GraphQLOverHTTP.md
```diff
@@ -500,9 +501,9 @@ execution - see "partial response" in the GraphQL specification.
 The server SHOULD NOT use a `4xx` or `5xx` status code.
 
 Note: For compatibility with legacy servers, this specification allows the use
-of `4xx` or `5xx` status code for failed requests using the `application/json`
-media type, but it is strongly discouraged. To use `4xx` and `5xx` status codes,
-please use the `application/graphql+json` media type.
+of `4xx` or `5xx` status code for failed requests where the responses uses the
```
ghmcadams, line 504: Consider `requests where the response uses` or `requests where the responses use`

spec/GraphQLOverHTTP.md
```diff
@@ -178,49 +227,74 @@ A request for execution should contain the following request parameters:
...
+For HTTP GET requests, the GraphQL request parameters MUST be provided in the
+query component of the request URL in the `application/x-www-form-urlencoded`
+format as specified by
+[WhatWG's URLSearchParams class](https://url.spec.whatwg.org/#interface-urlsearchparams).
+
+The `query` parameter MUST be the string representation of the Source Text of
+the Document as specified in
+[the Language section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Language).
+
+The `variables` parameter, if used, MUST be represented as a URL-encoded JSON
```
glasser, line 283: Why not specify the same thing for `extensions`?

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
```
glasser, line 311: (Previous line) Why not mention how `operationName` is encoded outside of the Note?

spec/GraphQLOverHTTP.md
```diff
@@ -178,49 +227,74 @@ A request for execution should contain the following request parameters:
...
-well established.
+Note: When comparing {GraphQL-over-HTTP request} against the term
+["request"](https://spec.graphql.org/draft/#request) in the GraphQL
+specification you should note the {GraphQL schema} and "initial value" are not
+included in the GraphQL-over-HTTP {request}, they are handled by the {server}
+based on the URL used.
+
+Note: Be aware that `query` is a misleading parameter name as its value is a
+string describing one or more operations, each of which may be a query, mutation
+or subscription. A better name would have been `document`, but the term `query`
```
glasser, line 238: This is the only reference to subscriptions in the spec. Perhaps it should be dropped, and/or there should be an explicit note that this spec doesn't provide a transport for subscriptions at this time.

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
...
+servers time to move over, it is not required to be supported until 1st
+January 2025.
+
+### Legacy watershed
+
+From 1st January 2025 (`2025-01-01T00:00:00Z`), a server MUST support requests
+which accept the `application/graphql+json` media type (as indicated by the
+`Accept` header).
+
+Before 1st January 2025 (`2025-01-01T00:00:00Z`), if the client does not supply
```
glasser, line 446: Is the implication here that spec-compatible servers should literally contain a conditional checking the current time (until we're in 2025 and that conditional can be removed)? Or I guess because this is merely a SHOULD it is OK for a server author to just decide to transition their code from one default to the other at a point that feels reasonable?

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
...
+If the GraphQL response contains a non-null {data} entry then the server MUST
+use the `200` status code.
+
+Note: This indicates that no {GraphQL request error} was raised, though one or
+more {GraphQL field error} may have been raised this is still a successful
+execution - see "partial response" in the GraphQL specification.
+
+The server SHOULD NOT use a `4xx` or `5xx` status code.
+
+Note: For compatibility with legacy servers, this specification allows the use
```
glasser, line 503: For what it's worth, Apollo Server does use 400s for GraphQL request errors (in a somewhat inconsistent way in AS3, and in a more consistent way in AS4). I don't think we are going to choose to change those to 200s for `application/json` as encouraged by this spec.  I do look forward to implementing `application/graphql+json` responses, though. My understanding is that this only violates SHOULDs, and I don't disagree with the recommendations of the spec.

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
...
+[differentiates field errors from request errors](https://spec.graphql.org/draft/#sec-Handling-Field-Errors)
+and refers to the situation wherein a {GraphQL field error} occurs as a partial
+response; it still indicates successful execution.
 
-### Client is not allowed to access the schema
+## Processing the response
 
-In case the client can not access the schema at all, the server SHOULD respond
-with the appropriate `4xx` status code.
+If the response uses a non-`200` status code and the media type of the response
```
glasser, line 610: For what it's worth, the Apollo implementation of persisted queries does violate this MUST NOT: persisted query-related errors are sent (currently) as `application/json` with 400s and are parsed by the client.

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
...
+#### Examples
 
-- [GraphQL specification validation](https://spec.graphql.org/October2021/#sec-Validation)
-- custom validation, for example: depth limit, complexity limit
+The following examples provide guidance on how to deal with specific error cases
+when using the `application/graphql+json` media type to encode the response
+body:
 
-The server SHOULD deny execution with a status code of `400` (Bad Request).
+##### Invalid request body or parsing failure
```
glasser, line 555: Should this also include failure to parse `variables` or `extensions` provided as GET parameters?

It looks like this is about failures to produce a GraphQL request. But that's separate from GraphQL parse failures... which should probably have its own section?

glasser (commented):
I think Apollo Server is relatively close to obeying the MUSTs in this document. The main way it doesn't is by allowing `query`-less operations for APQ support.

(Additionally, Apollo Server also supports a "batched" protocol not described here where the body can be an array of requests rather than a single request. This protocol has its drawbacks — eg, ideally responses could be streamed via multi-part instead of returned as one big JSON array. The main reason people like this protocol is that it allows request setup (such as context value calculation) to occur once for all operations in the batch (sharing resources like DataLoaders across operations) rather than separately. However, supporting this additional syntax doesn't necessarily put AS out of spec, I think.)

spec/GraphQLOverHTTP.md
benjie: > The Apollo implementation of automatic persisted queries works by leaving out query in the particular requests [...] If we want to try to make the Apollo APQ protocol not a violation of this spec, we could make query optional and have some sort of note stating that when query is left out, the server must have some sort of extensions-based mechanism for deriving the Document.

I think it's okay for Automatic Persisted Queries to not be an official GraphQL request - the point of the specification is to define a common set of rules that means a compliant client from any vendor can speak to a compliant server from any vendor. APQ is not part of that flow, it's a specific behavior that the client/server can negotiate as an extension, separate from this specification.

> Does the spec actually explain anywhere what to do if the query is not provided? I suppose for GETs this falls into the "Server URLs which enable GraphQL requests MAY also be used for other purposes" thing to some degree.

"A {GraphQL-over-HTTP request} is formed of the following parameters" implies that missing "query" means that it's not a "GraphQL over HTTP request"; then we have "When a server receives a {GraphQL-over-HTTP request}, it must return a well‐formed response." but it's not a GraphQL over HTTP request, so it's up to the server how to handle it, either by classifying it as "other purposes" (in this case serving GraphQL but in a way not 100% compatible with the GraphQL-over-HTTP spec), or it can throw an error.

Maybe we could add a non-normative note to make this clearer? Or indicate that there may be more ways to talk GraphQL with a webserver than strictly defined in this spec?

spec/GraphQLOverHTTP.md
```diff
@@ -178,49 +227,74 @@ A request for execution should contain the following request parameters:
 - {extensions} - (_Optional_): This entry is reserved for implementors to extend
   the protocol however they see fit.
 
-Note: Be aware that `query` is a misleading name as it can contain a string
-describing multiple operations, each of which may be a query, mutation or
-subscription. A better name would have been `document`, but the term `query` is
-well established.
+Note: When comparing {GraphQL-over-HTTP request} against the term
```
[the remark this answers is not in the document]
benjie: Please raise an editorial PR, it's not clear to me the concern you have here.

spec/GraphQLOverHTTP.md
benjie: I've never noticed extensions being used in a URL, but you noted it is used for APQ so I'll add it :+1:

spec/GraphQLOverHTTP.md
benjie: 1. Should it contain a conditional check on the time? Yes, ideally.
2. Can it just decide to transition the code from one default to another at a point that feels reasonable? Yes, knowing that it could cause issues for legacy clients.

Note: client's SHOULD include `Accept` and SHOULD accept `application/graphql+json`, so the only clients affected by this will be clients who a) do not issue `Accept` headers, and b) actually bother to read the `Content-Type` of the response and explicitly do not support `application/graphql+json`. I wonder what the size of that Venn diagram overlap is - perhaps you have data on that at Apollo?

I wonder if we should make it so that clients MUST include an `Accept` header from the watershed?

spec/GraphQLOverHTTP.md
benjie: Not changing them is fine. Once 2025 rolls around this'll be a non-issue hopefully :crossed_fingers:

spec/GraphQLOverHTTP.md
benjie: I've tried editing this in, but it just gets more confusing because there's so many different ways of encoding a bad request. Since it's just an example I think it's fine to leave it as-is. What do you think?

spec/GraphQLOverHTTP.md
glasser: >I think it's okay for Automatic Persisted Queries to not be an official GraphQL request - the point of the specification is to define a common set of rules that means a compliant client from any vendor can speak to a compliant server from any vendor. APQ is not part of that flow, it's a specific behavior that the client/server can negotiate as an extension, separate from this specification.

Yep, that's fair. I just wanted to clarify that with our "automatic" PQ implementation, you don't have "a server that only allows persisted operations".

Re the second point, no strong feelings here (I think the spec is fine as given).

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
glasser: https://github.com/graphql/graphql-over-http/pull/199/files

spec/GraphQLOverHTTP.md
glasser: I think there's a pretty big difference between "you did a GraphQL-over-HTTP-JSON-level error by failing to post good JSON" and "you did a GraphQL-level error by failing to post good GraphQL".  It seems like you are explicitly showing how every other step in the execution process is handled (validation, operation resolution, variable coercion) so not including GraphQL parsing is a notable absence. I'm just talking something like `{"query": "}"}`.

spec/GraphQLOverHTTP.md
benjie: Yeah; curl et al is the main reason that I left this as a SHOULD in the first place; but it's also why I figure if you don't supply `Accept` you probably don't care about what the server says the `Content-Type` is :laughing:

spec/GraphQLOverHTTP.md
benjie: Fair. Let's split invalid parameters and parsing failure into two separate examples :+1:  - I'll raise a PR.

spec/GraphQLOverHTTP.md
ghmcadams: > Should it contain a conditional check on the time? Yes, ideally.

I see this differently.  I don't see the purpose of a watershed as to set a precise date at which functionality should be turned on.  The way I see it, a watershed is to set a date at which a server without such functionality is no longer compliant.  During the time between now and the date of the watershed, new releases of a server should gradually start to support the new functionality described in the spec.  Ideally (and totally at the discretion of the implementor), a server would be backward compatible.

benjie: I broadly agree, but David has pointed out the point where there is a sudden change in the recommended behaviour - the interpretation of a request without an `Accept` header should be `application/json` before and `application/graphql+json` after. That said, since it's only a SHOULD it doesn't matter too much and you don't _need_ a time check.

spec/GraphQLOverHTTP.md
```diff
@@ -237,120 +311,304 @@ omitted.
...
+If [GetOperation()](<https://spec.graphql.org/draft/#GetOperation()>) raises a
+{GraphQL request error}, the server SHOULD NOT execute the request and SHOULD
+return a status code of `400` (Bad Request).
+
+##### Variable coercion failure
+
+If
+[CoerceVariableValues()](<https://spec.graphql.org/draft/#CoerceVariableValues()>)
+raises a {GraphQL request error}, the server SHOULD NOT execute the request and
+SHOULD return a status code of `400` (Bad Request).
```
glasser, line 587: One note: while this is a good idea which I support (and which Apollo Server tries to implement), it's challenging to implement this today in a non-hacky way with `graphql-js`. I thought there was already an issue for this but I couldn't find one so I filed https://github.com/graphql/graphql-js/issues/3679

benjie: Yeah; I believe the trick right now is that if it doesn't have a `data` property (note that's different from it being present but null) then you can assume it's a `CoerceVariableValues()` or similar request error as opposed to a field error. I think you can actually just straight up use `graphql(...)` and use this heuristic to adhere to the GraphQL-over-HTTP spec, you don't need to call out to the separate methods since all the request errors end up in HTTP 400 anyway. But yeah, it'd be beneficial to have it separate for sure :+1:

glasser:
Ah true, I believe we do map lack of data to 400 but we also want a more
precise error code inline for the different error types and that's what we
can't do without better support from graphql-js.

spec/GraphQLOverHTTP.md
glasser: Ah true, I believe we do map lack of data to 400 but we also want a more
precise error code inline for the different error types and that's what we
can't do without better support from graphql-js.

benjie: 💯

Issue #191, opened 2022-07-11
[2022-06-27] Incorporate David's and working group's feedback


benjie:
In particular:

- “If you use other media types, make sure you’re aware of the security consequences of doing so”
- We should add a note explaining why servers should reject a request with no content-type header, but perhaps not forbid it outright.
- Any time we have a should, we should add a non-normative note explaining why

PR #197, opened 2022-07-11
Add "Note" after various "SHOULD" rules in the spec

benjie:
I've not covered all of them, but hopefully enough to be helpful.

Part of #191

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -239,8 +239,16 @@ A client SHOULD indicate the media types that it supports in responses using the
 `Accept` HTTP header as specified in
 [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
 
+Note: If a client does not supply the `Accept` header then the server may
+respond with an error, or with any content type it chooses. To ensure you client
+gets something useful, it should indicate the media types it supports.
+
 If the client supplies an `Accept` header, the client SHOULD include the media
 type `application/graphql+json` in the `Accept` header.
+
+Note: From 1st Jan 2025, every _server_ and _client_ must support
+`application/graphql+json`, so including this in the Accept header should give
+your client compatibility with any _server_.
 
 ### Legacy Watershed
 
@@ -316,11 +324,17 @@ type (as indicated by the `Content-Type` header).
 If the client does not supply a `Content-Type` header with a POST request, the
 server SHOULD reject the request using the appropriate `4xx` status code.
 
+Note: This is left as a "should" to allow your server to choose a different
+behavior should it wish to.
+
 A server MAY support POST requests encoded with and/or accepting other media
 types.
 
 If a client does not know the media types the server supports then it SHOULD
 encode the request body in JSON (i.e. with `Content-Type: application/json`).
+
+Note: Request encoding with media type `application/json` is supported by every
+compliant _server_.
 
 ### JSON Encoding
 
@@ -457,6 +471,9 @@ GraphQL response, the server SHOULD respond with the appropriate status code
 depending on the concrete error condition.
 
 Note: Typically this will be the `400` (Bad Request) status code.
+
+Note: This rule is "should" to maintain compatibility with legacy servers which
+may return 200 status codes even when this type of error occurs.
 
 Otherwise, the status codes depends on the media type with which the GraphQL
 response will be served:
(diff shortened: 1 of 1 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -236,9 +239,17 @@ A client SHOULD indicate the media types that it supports in responses using the
 `Accept` HTTP header as specified in
 [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
 
+Note: If a client does not supply the `Accept` header then the server may
+respond with an error, or with any content type it chooses. To ensure you client
```
ghmcadams, line 243: Typo?

```suggestion
respond with an error, or with any content type it chooses. To ensure your client
```

spec/GraphQLOverHTTP.md
```diff
@@ -313,12 +324,18 @@ type (as indicated by the `Content-Type` header).
 If the client does not supply a `Content-Type` header with a POST request, the
 server SHOULD reject the request using the appropriate `4xx` status code.
 
+Note: This is left as a "should" to allow your server to choose a different
+behavior should it wish to.
```
ghmcadams, line 328: Is this needed?  As worded, it is no more than a casual definition of `SHOULD`.  If the intent is to reiterate and remind readers that a server can choose to vary behavior when a `SHOULD` is used, consider something like this:

```suggestion
Note: Rejecting such requests encourages clients to supply a `Content-Type` header with every POST request.  A server has the option to assume any media type they wish when none is supplied, with the understanding that parsing the request may fail.
```

spec/GraphQLOverHTTP.md
```diff
@@ -313,12 +324,18 @@ type (as indicated by the `Content-Type` header).
...
+
 A server MAY support POST requests encoded with and/or accepting other media
 types.
 
 If a client does not know the media types the server supports then it SHOULD
 encode the request body in JSON (i.e. with `Content-Type: application/json`).
 
+Note: Request encoding with media type `application/json` is supported by every
+compliant _server_.
```
ghmcadams, line 337: minor: My opinion is that italicizing keywords should be used always or never.

spec/GraphQLOverHTTP.md
```diff
@@ -455,6 +472,9 @@ depending on the concrete error condition.
 
 Note: Typically this will be the `400` (Bad Request) status code.
 
+Note: This rule is "should" to maintain compatibility with legacy servers which
+may return 200 status codes even when this type of error occurs.
```
ghmcadams, line 476: This note implies a changing normative rule that doesn't exist.  Consider rewording.

spec/GraphQLOverHTTP.md
benjie: The italics actually serves to turn the word into a [spec-md reference](https://spec-md.com/#sec-Definition-References), in this case clarifying that "server" definitely refers to "GraphQL-over-HTTP compliant server" or whatever we defined it to be.

benjie: (But I would be perfectly okay with every appropriate mention of client/server being a reference.)

spec/GraphQLOverHTTP.md
benjie, line 476:
```suggestion
can return 200 status codes even when this type of error occurs.
```

spec/GraphQLOverHTTP.md
```diff
@@ -455,6 +474,9 @@ depending on the concrete error condition.
 
 Note: Typically this will be the `400` (Bad Request) status code.
 
+Note: This rule is "should" to maintain compatibility with legacy servers which
```
spawnia, line 477:
```suggestion
Note: This rule is "SHOULD" to maintain compatibility with legacy servers which
```

spec/GraphQLOverHTTP.md
```diff
@@ -166,6 +166,9 @@ The GraphQL specification allows for many
 Servers and clients MUST support JSON and MAY support other, additional
 serialization formats.
 
+Note: If you allow other media types, particularly on requests, please ensure
+that you are aware of the security consequences of doing so.
```
spawnia, line 170:
```suggestion
Note: Allowing other media types, particularly on requests, can be insecure.
```

benjie: It's more like "can be insecure if you don't take the necessary precautions/actions".

ghmcadams (approved):

spawnia (approved):

PR #198, opened 2022-07-11
Clarify operationName in the query string

benjie:
This is a normative change. Whilst adding the SHOULDs for #191 I realised that this should be a MUST, otherwise there's ambiguity in how `operationName=null` should be handled.

cc @glasser


ghmcadams:
This change makes sense.  It is my belief that it should be removed from a `Note` to make it normative.

benjie:
Ahhhh that explains the loose language before! I’ll revise, thanks for the spot 👍

ghmcadams (approved):

michaelstaib (approved):

Issue #183, opened 2022-07-07
Consider adding SSE for Subscription operations to the specification


n1ru4l:
Server-Sent Events (SSE) is a server push technology enabling a client to receive automatic updates from a server via an HTTP connection, that does not require any complex protocol compared to WebSocket, thus making it a perfect fit for having a recommended and standardized way of doing GraphQL Subscrptions in the GraphQL over HTTP specification

____

- https://github.com/enisdenjo/graphql-sse
- https://www.graphql-yoga.com/docs/features/subscriptions

benjie:
We definitely intend to do this (I've been speaking with @enisdenjo about it) but it is out of scope for the 1.0 release which we basically need to feature freeze right now in order to try and get it cut along with the GraphQL spec in October.

enisdenjo:
Currently a [GraphQL over SSE Protocol RFC](https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverSSE.md).

Even though SSE is indeed quite simple, there are some "complexities" added to the Protocol which aim to overcome common problems, just to name a few:
- The "single connection mode" aiming to help with HTTP/1 servers and subscription heavy apps
- The requirement of reporting errors through the established SSE connection instead of the HTTP response because of the [ultra vague `EventSource.onerror`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/error_event)
- The presence of a `complete` message to distinguish when a connection gets closed because of a network loss vs. when the subscription is actually completed

n1ru4l:
> The "single connection mode" aiming to help with HTTP/1 servers and subscription heavy apps
In my opinion, this added complexity can be optional as servers that support HTTP/2 don't need to do this. [HTTP/2 support for browsers seems quite high (97%)](https://caniuse.com/?search=http2). [From quick research, overall the current adoption for web servers is around 44%](https://w3techs.com/technologies/details/ce-http2)

enisdenjo:
> In my opinion, this added complexity can be optional as servers that support HTTP/2 don't need to do this.

[...]

PR #202, opened 2022-07-14
Add note stating Subscriptions are out of scope

benjie:
As suggested by @rivantsov in https://github.com/graphql/graphql-over-http/issues/183#issuecomment-1182795688

(Non-normative, merge if happy.)


n1ru4l (approved):
😁

PR #209, opened 2022-07-24
Organize existing agendas and notes by year

spawnia:


ghmcadams (approved):

benjie:
We didn't do this in the GraphQL-WG repo because it would break links that didn't use GitHub permalink format; sadly GitHub doesn't forward links to deleted files.

ghmcadams:
Oh No.  I didn't think about permalinks.  One positive is that we are fairly early... There are not a lot of agendas yet.  Maybe it's better we did this now and have a good organization moving forward.

PR #210, opened 2022-07-24
Reflect move to stage 1

spawnia:
As discussed in the July 2022 working group meeting,
we have now officially moved the spec to stage 1.


spawnia:
Should we update the GitHub description too?

image

ghmcadams (approved):

PR #211, opened 2022-07-24
Omit trailing slash from URLs

spawnia:
If the GitHub link could be updated as well we would be consistent throughout:

image


ghmcadams (approved):

benjie:
Just a note that `https://example.com/foo` and `https://example.com/foo/` are concretely different URLs. Many servers will auto-forward from one to the other, but you cannot assume that this is the case for all URLs. One notable exception to this is the root path `https://example.com` and `https://example.com/` which are equivalent for all websites.

I tested a few of these links and didn't find any that break, but it was not exhaustive.

ghmcadams:
Thanks for that, @benjie.

@spawnia could you verify that all URLs changed in this PR are still accurate and working?  If any are not, we can revert.

spawnia:
All URLs are fine, except melody.sh which is entirely offline - see https://github.com/graphql/graphql-over-http/pull/217

PR #224, opened 2022-09-06
Query string is URL-encoded

sungam3r:
Query parameter should also be URL-encoded (according to provided example).

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -272,7 +272,7 @@ query component of the request URL in the `application/x-www-form-urlencoded`
 format as specified by
 [WhatWG's URLSearchParams class](https://url.spec.whatwg.org/#interface-urlsearchparams).
 
-The `query` parameter MUST be the string representation of the Source Text of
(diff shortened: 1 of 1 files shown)


benjie:
I don't believe this change is needed, nor desired. We've already stated that the parameters must be given according to the URLSearchParams class, which implicitly means they're URL encoded.

If we adopt this proposed change ("The `query` parameter MUST be the URL-encoded string representation of the Source Text") then someone might think they need to do this:

```js
const p = new URLSearchParams();
p.append('query', encodeURIComponent('query { myField(arg: "&+") }'));
```

It would result in the query being double-encoded:

```js
p.toString();
// 'query=query%2520%257B%2520myField%28arg%253A%2520%2522%2526%252B%2522%29%2520%257D'
```

Shane32:
> I don't believe this change is needed, nor desired. We've already stated that the parameters must be given according to the URLSearchParams class, which implicitly means they're URL encoded.

I agree.  The parameter value itself is not encoded but rather the _set_ of parameters are encoded in a specific format, and this is explicitly stated in the preceding paragraph.

Issue #230, opened 2022-10-12
What is well-formed response


TomokiMiyauci:
> When a server receives a GraphQL-over-HTTP request, it must return a well-formed response.

> In case of errors that completely prevent the generation of a well-formed GraphQL response, the server SHOULD respond with the appropriate status code depending on the concrete error condition.

What is the difference between a well-formed response and a well-formed GraphQL response?

As I understand it, a Well-formed GraphQL response is a JSON serialization with `data`, `errors` or `extensions` as keys. 

For example, if the client's "Accept" header is `text/*`, it has chosen to return a `406` status code.

Should the response body in this case be a JSON serialization with `errros` as the key?

spawnia:
I see how the specification is unclear in its current form, the link for `GraphQL Response` should point to another URL. See https://github.com/graphql/graphql-over-http/pull/231#discussion_r993170327.

ghmcadams:
@TomokiMiyauci Thank you for raising this issue.  #231 has been merged.  Do you believe it resolves this issue?

TomokiMiyauci:
@ghmcadams Resolved. Thank you.

PR #231, opened 2022-10-12
Clarify well-formed response and fix links to GraphQL response spec

spawnia:
Resolves https://github.com/graphql/graphql-over-http/issues/230

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -183,7 +183,7 @@ using the JSON encoding for GraphQL requests:
 | ------------------ | --------------------------------------- |
 | `application/json` | Standard type for GraphQL JSON requests |
 
-And for _GraphQL responses_:
+And for a _GraphQL response_:
 
 | Name                                | Description                                                        |
 | ----------------------------------- | ------------------------------------------------------------------ |
@@ -531,7 +531,7 @@ status code.
 Note: The result of executing a GraphQL operation may contain partial data as
 well as encountered errors. Errors that happen during execution of the GraphQL
 operation typically become part of the result, as long as the server is still
-able to produce a well-formed response. There's currently not an approved HTTP
+able to produce a well-formed _GraphQL response_. There's currently not an approved HTTP
 status code to use for a "partial response," contenders include WebDAV's status
 code "207 Multi-Status" and using a custom code such as "247 Partial Success."
 [IETF RFC2616 Section 6.1.1](https://datatracker.ietf.org/doc/html/rfc2616#section-6.1.1)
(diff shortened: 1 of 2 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -401,8 +401,8 @@ And the body:
 
 # Response
 
-When a server receives a _GraphQL-over-HTTP request_, it must return a
-well‐formed response. The server's response describes the result of validating
+When a server receives a well-formed _GraphQL-over-HTTP request_, it must return a
+well‐formed [GraphQL response](#sec-Response). The server's response describes the result of validating
```
spawnia, line 405: I think this is supposed to link to https://spec.graphql.org/October2021/#sec-Response - however that does not seem to work correctly in the rendered output:

image

How can we correct this?

benjie: By italicising we can leverage spec-md's referencing feature to reference the spec:

```suggestion
well‐formed _GraphQL response_. The server's response describes the result of validating
```

Enabled by:

https://github.com/graphql/graphql-over-http/blob/466f5db37eac577fe999d9e8118317919b86385a/spec/metadata.json#L7

benjie (changes requested):
I believe the intent you have here is correct 👍 

Please also apply this fix to https://github.com/graphql/graphql-over-http/blob/466f5db37eac577fe999d9e8118317919b86385a/spec/GraphQLOverHTTP.md?plain=1#L534

spec/GraphQLOverHTTP.md
spawnia: Thank you, I corrected the remaining references.

spec/GraphQLOverHTTP.md
```diff
@@ -183,7 +183,7 @@ using the JSON encoding for GraphQL requests:
 | ------------------ | --------------------------------------- |
 | `application/json` | Standard type for GraphQL JSON requests |
 
-And for GraphQL responses:
+And for _GraphQL responses_:
```
benjie, line 186: The GraphQL spec always uses the singular for keywords (probably because of `spec-md` behaviors, but also to make matching/indexing easier), I suggest we stick with this pattern too:

```suggestion
And for a _GraphQL response_:
```

spec/metadata.json
```diff
@@ -5,6 +5,7 @@
       "graphql-schema": "#sec-Schema",
       "graphql-request": "#request",
       "graphql-response": "#sec-Response",
+      "graphql-responses": "#sec-Response",
```
benjie, line 8: Please remove :)

benjie (changes requested):
I asked Lee before if every instance of keyword should be `_keyword_` and he said he tends to only do the first in any paragraph/section. That said, this is fine and I think having too many links is probably better than having too few 👍

benjie (approved):
Thanks for your work on this!

benjie:
This is effectively an editorial change, so whoever reviews this next if you have permissions and approve it, please go ahead and merge it 👍

ghmcadams (approved):

PR #240, opened 2023-01-27
Update nodemon

benjie:


benjie:
Also: I accidentally pushed to `main` an upgrade to spec-md to solve this bug: https://github.com/leebyron/spec-md/pull/92

PR #241, opened 2023-02-14
`application/json` should only mandate 200 for _well-formed_ GraphQL-over-HTTP requests

benjie:
Reviewing @enisdenjo's GraphQL-over-HTTP audits, I realise that I made a mistake when drafting this specification. In my head, each time I wrote "request" I meant "GraphQL-over-HTTP request", i.e. one that roughly conforms to:

```ts
type GraphQLOverHTTPRequest = {
  query: string;
  operationName?: string | null;
  variables?: Record<string, any> | null;
  extensions?: Record<string, any> | null;
}
```

However this was not explicitly stated in enough places, the end result being that we now effectively state that if you send garbage at a GraphQL legacy server it _should_ return a 200 back at you. This was not my intent - my intent that 200 should only be used if the request conformed to the TypeScript shape above - `query` doesn't have to be parseable, and doesn't have to validate against your schema, but it does have to be present and it does have to be a string. If you can't even parse the JSON body of a request, you should definitely be returning a `400 Bad Request` not a `200 Okay`.

I've performed some edits to make this clearer, and have also copied the examples from `application/graphql-response+json` to `application/json` and corrected them for that location. I've also moved a few notes around where they would have been repeated otherwise, and made a few other small edits in line with accomplishing the above.

cc @graphql/http-maintainers

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -208,7 +208,7 @@ GET.
 :: A _GraphQL-over-HTTP request_ is an HTTP request that encodes the following
 parameters in one of the manners described in this specification:
 
-- {query} - (\_Required\_, string): The string representation of the Source Text
+- {query} - (_Required_, string): The string representation of the Source Text
   of a GraphQL Document as specified in
   [the Language section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Language).
 - {operationName} - (_Optional_, string): The name of the Operation in the
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -571,7 +571,7 @@ when using the `application/json` media type to encode the response body:
 For example a POST request body of `NONSENSE` or `{"query":` (note: invalid
 JSON).
 
-Requests that the server cannot interpret should result in status code `400`
+Requests that the server cannot interpret SHOULD result in status code `400`
 (Bad Request).
 
 ##### Invalid parameters
@@ -580,14 +580,14 @@ For example a POST request body of `{"qeury": "{__typename}"}` (note: typo) or
 `{"query": "query Q ($i:Int!) { q(i: $i) }", "variables": [7]}` (note: invalid
 shape for `variables`).
 
-A requests that does not constitute a well-formed _GraphQL-over-HTTP request_
-should result in status code `400` (Bad Request).
+A request that does not constitute a well-formed _GraphQL-over-HTTP request_
+SHOULD result in status code `400` (Bad Request).
 
 ##### Document parsing failure
 
 For example a POST request body of `{"query": "{"}`.
 
-Requests where the _GraphQL document_ cannot be parsed should result in status
+Requests where the _GraphQL document_ cannot be parsed SHOULD result in status
 code `200` (Okay).
 
 ##### Document validation failure
@@ -685,7 +685,7 @@ For example a POST request body of `{"qeury": "{__typename}"}` (note: typo) or
 shape for `variables`).
 
 A request that does not constitute a well-formed _GraphQL-over-HTTP request_
-should result in status code `400` (Bad Request).
+SHOULD result in status code `400` (Bad Request).
 
 ##### Document parsing failure
 
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -279,18 +279,21 @@ provided in the query component of the request URL, encoded in the
 `application/x-www-form-urlencoded` format as specified by the
 [WhatWG URLSearchParams class](https://url.spec.whatwg.org/#interface-urlsearchparams).
 
-The {query} parameter MUST be the URL-encoded string representation of the
-source text of the document as specified in
+The {query} parameter MUST be the string representation of the source text of
+the document as specified in
 [the Language section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Language).
 
-The {operationName} parameter, if present, must be a URL-encoded string.
+The {operationName} parameter, if present, must be a string.
 
 Each of the {variables} and {extensions} parameters, if used, MUST be encoded as
-a URL-encoded JSON string.
+a JSON string.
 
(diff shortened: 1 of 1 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -267,25 +274,26 @@ still leveraging modern features if available in the server.
...
-query component of the request URL, encoded in the
+For HTTP GET requests, the _GraphQL-over-HTTP request_ parameters MUST be
+provided in the query component of the request URL, encoded in the
 `application/x-www-form-urlencoded` format as specified by the
 [WhatWG URLSearchParams class](https://url.spec.whatwg.org/#interface-urlsearchparams).
 
-The `query` parameter MUST be the string representation of the Source Text of
-the Document as specified in
+The {query} parameter MUST be the URL-encoded string representation of the
+Source Text of the Document as specified in
```
spawnia, line 283:
```suggestion
source text of the document as specified in
```

The linked specification also writes them in lower case

spec/GraphQLOverHTTP.md
```diff
@@ -511,13 +532,86 @@ Note: This indicates that no _GraphQL request error_ was raised, though one or
 more _GraphQL field error_ may have been raised this is still a successful
 execution - see "partial response" in the GraphQL specification.
 
-The server SHOULD NOT use a `4xx` or `5xx` status code.
+The server SHOULD NOT use a `4xx` or `5xx` status code a response to a
```
spawnia, line 535:
```suggestion
The server SHOULD NOT use a `4xx` or `5xx` status code for a response to a
```

spawnia (commented):
Good change overall

spawnia (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -267,25 +274,26 @@ still leveraging modern features if available in the server.
...
-For HTTP GET requests, the GraphQL request parameters MUST be provided in the
-query component of the request URL, encoded in the
+For HTTP GET requests, the _GraphQL-over-HTTP request_ parameters MUST be
+provided in the query component of the request URL, encoded in the
 `application/x-www-form-urlencoded` format as specified by the
 [WhatWG URLSearchParams class](https://url.spec.whatwg.org/#interface-urlsearchparams).
 
-The `query` parameter MUST be the string representation of the Source Text of
-the Document as specified in
+The {query} parameter MUST be the URL-encoded string representation of the
```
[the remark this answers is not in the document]
benjie: Nope, you (and Past Benjie) are right. I’m not 100% certain we’re defining things in the right way; we might need a “decoding” step to be explicit, because there’s ambiguity around e.g. `variables` the GraphQL-over-HTTP request parameter (map) and `variables` the query parameter (JSON encoded string). I might need to introduce a different clearer term.

enisdenjo (approved):
This is great, often GQL servers are in situations where they don't do the parsing themselves (fastify or express does it for you) and therefore have limited control.

Also related to https://github.com/graphql/graphql-http/pull/47.

benjie:
@Shane32 I've updated the text you commented on and added a `Note:` - do you think this adds clarity, or the `Note:` should be omitted?

Shane32:
> @Shane32 I've updated the text you commented on and added a `Note:` - do you think this adds clarity, or the `Note:` should be omitted?

I don’t mind the extra clarity. Personally, it’s second nature to me to always ensure query arguments and sql parameters are correctly encoded ever since the days of Classic ASP, but judging from the amount of sql injection attacks over the years, a little extra clarity here is probably a good thing.

FYI, another option is to revert the change, and instead add a note stating that parameters are not double encoded but were restated for clarity.

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
benjie: This has been addressed via a note.

spec/GraphQLOverHTTP.md
```diff
@@ -517,13 +541,86 @@ Note: This indicates that no _GraphQL request error_ was raised, though one or
...
+(Bad Request).
+
+##### Invalid parameters
+
+For example a POST request body of `{"qeury": "{__typename}"}` (note: typo) or
+`{"query": "query Q ($i:Int!) { q(i: $i) }", "variables": [7]}` (note: invalid
+shape for `variables`).
+
+A requests that does not constitute a well-formed _GraphQL-over-HTTP request_
+should result in status code `400` (Bad Request).
```
Shane32, line 584:
```suggestion
SHOULD result in status code `400` (Bad Request).
```

spec/GraphQLOverHTTP.md
```diff
@@ -517,13 +541,86 @@ Note: This indicates that no _GraphQL request error_ was raised, though one or
...
+
+The following examples provide guidance on how to deal with specific error cases
+when using the `application/json` media type to encode the response body:
+
+##### JSON parsing failure
+
+For example a POST request body of `NONSENSE` or `{"query":` (note: invalid
+JSON).
+
+Requests that the server cannot interpret should result in status code `400`
```
Shane32, line 574:
```suggestion
Requests that the server cannot interpret SHOULD result in status code `400`
```

spec/GraphQLOverHTTP.md
```diff
@@ -587,8 +684,8 @@ For example a POST request body of `{"qeury": "{__typename}"}` (note: typo) or
 `{"query": "query Q ($i:Int!) { q(i: $i) }", "variables": [7]}` (note: invalid
 shape for `variables`).
 
-Requests that do not constitute a well-formed _GraphQL-over-HTTP request_ should
-result in status code `400` (Bad Request).
+A request that does not constitute a well-formed _GraphQL-over-HTTP request_
+should result in status code `400` (Bad Request).
```
Shane32, line 688:
```suggestion
SHOULD result in status code `400` (Bad Request).
```

Shane32: lines 678 and 694 also

enisdenjo:
Changes to the audit suite are already merged in graphql-http https://github.com/graphql/graphql-http/pull/62.

PR #252, opened 2023-05-15
Incremental Delivery over HTTP/2

enisdenjo:
HTTP/2 provides better mechanisms of data streaming and therefore prohibits the use of the Transfer-Encoding header (which is [recommended for GraphQL's Incremental Delivery](https://github.com/graphql/graphql-over-http/blob/main/rfcs/IncrementalDelivery.md#transfer-encoding-chunked)).

This PR explains the situation in greater detail and defines behaviour for HTTP/2 servers.

### DIFF
--- a/rfcs/IncrementalDelivery.md
+++ b/rfcs/IncrementalDelivery.md
@@ -14,7 +14,7 @@ An incrementally delivered response should contain the `Transfer-Encoding: chunk
 
 Because of improved data streaming mechanisms, HTTP/2 prohibits the use of the `Transfer-Encoding` header. It is very likely that compliant servers will treat requests containing the header as malformed ([see section 8.2.2. Connection-Specific Header Fields in HTTP/2 spec](https://datatracker.ietf.org/doc/html/rfc9113#section-8.1)).
 
-Having said this, HTTP/2 enabled servers must not set the `Transfer-Encoding` header during response.
+Compliant servers MUST follow the HTTP/2 specification and not set the `Transfer-Encoding` header.
 
 ## `content-type: multipart/mixed`
 
--- a/rfcs/IncrementalDelivery.md
+++ b/rfcs/IncrementalDelivery.md
@@ -14,7 +14,7 @@ An incrementally delivered response should contain the `Transfer-Encoding: chunk
 
 Because of improved data streaming mechanisms, HTTP/2 prohibits the use of the `Transfer-Encoding` header. It is very likely that compliant servers will treat requests containing the header as malformed ([see section 8.2.2. Connection-Specific Header Fields in HTTP/2 spec](https://datatracker.ietf.org/doc/html/rfc9113#section-8.1)).
 
-Compliant servers MUST follow the HTTP/2 specification and not set the `Transfer-Encoding` header.
+Compliant servers must follow the HTTP/2 specification and not set the `Transfer-Encoding` header.
 
 ## `Content-Type: multipart/mixed`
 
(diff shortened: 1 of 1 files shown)


rfcs/IncrementalDelivery.md
```diff
@@ -6,9 +6,15 @@ This RFC proposes adding a specification of how "Incremental" GraphQL results sh
 
 An Incremental result is a GraphQL result that is split up into multiple payloads, allowing clients quicker access to parts of the results. Currently this is supported by the proposed `@defer` and `@stream` directives ([RFC](https://github.com/graphql/graphql-spec/blob/master/rfcs/DeferStream.md)).
 
-## `transfer-encoding: chunked`
+### HTTP/1.1
 
-The HTTP response for an incrementally delivered response should contain the `transfer-encoding: chunked` response header. Chunked transfer encoding allows the body of the response to be delivered as a series of chunks, allowing clients to read each chunk of the response as it is sent without waiting for the entire response.
+An incrementally delivered response should contain the `transfer-encoding: chunked` response header when using HTTP/1.1. Chunked transfer encoding allows the body of the response to be delivered as a series of chunks, allowing clients to read each chunk of the response as it is sent without waiting for the entire response.
```
spawnia, line 11: I think we should always capitalize HTTP headers `Like-This: value` in this entire document.

rfcs/IncrementalDelivery.md
```diff
@@ -6,9 +6,15 @@ This RFC proposes adding a specification of how "Incremental" GraphQL results sh
...
+
+### HTTP/2
+
+Because of improved data streaming mechanisms, HTTP/2 prohibits the use of the `transfer-encoding` header. It is very likely that compliant servers will treat requests containing the header as malformed ([see section 8.2.2. Connection-Specific Header Fields in HTTP/2 spec](https://datatracker.ietf.org/doc/html/rfc9113#section-8.1)).
+
+Having said this, HTTP/2 enabled servers must not set the `transfer-encoding` header during response.
```
spawnia, line 17: This seems redundant with the first sentence of the above paragraph.

> HTTP/2 prohibits the use of the `transfer-encoding` header.

spawnia (changes requested):

rfcs/IncrementalDelivery.md
enisdenjo: The paragraph above is an explanation/description and this is the spec requirement. I added it for brevity.

rfcs/IncrementalDelivery.md
enisdenjo: I was matching the style to keep the changeset at the minimum - I can follow-up with a PR to take care of styling issues.

spawnia (changes requested):

rfcs/IncrementalDelivery.md
enisdenjo: I like the idea, thanks! https://github.com/graphql/graphql-over-http/pull/252/commits/b3ebf9f47fb3e2bf7bfdd9352cbeb58cbe54792e

maraisr:
what do we do instead?

enisdenjo:
sorry, I dont understand your question @maraisr

do instead of what? which parts dont make sense to you?

maraisr:
you're suggesting a preference for http/2 am i correct?

enisdenjo:
> you're suggesting a preference for http/2 am i correct? 

no, there's still a section describing what should happen on http/1.1. I am simply trying to eliminate any possible confusion by recommending a header thats prohibited in http/2

> and if so, and you remove the chunk header, doesn't that break the streaming aspect for browsers?

it does not

maraisr:
okay great! so no need to delete meros, still useful! 😌

maraisr:
@enisdenjo what should users do in http/2 to support incremental delivery provided the header is prohibited? Does this mean there is no support for defer/stream in http/2?

spawnia (approved):

enisdenjo:
> what should users do in http/2 to support incremental delivery provided the header is prohibited?

they should do what the spec says, not set the `Transfer-Encoding` header

> Does this mean there is no support for defer/stream in http/2?

it does not

@maraisr I hope https://github.com/enisdenjo/http2-multipart-mixed gives you a peace of mind.

PR #253, opened 2023-08-08
RFC: persisted operations

JoviDeCroock:
This RFC intends to add `(Automatic) Persisted Operations` to the GraphQL over HTTP spec.

With persisted operations we can cater to some common needs

- Caching (determinstic document hash and url-size)
- Security (we can choose which requests to allow)

This requires us to agree on a way to `stringify` the document, a way to `hash` the document and an alteration to our current request parameters as this tries to add `documentId` and make `query` optional.

### DIFF
--- a/rfcs/PersistedOperations.md
+++ b/rfcs/PersistedOperations.md
@@ -12,15 +12,15 @@ With Persisted Operations we have a few goals:
 
 ## Flow
 
-### Producing the hash
+### Producing the document-id
 
-To produce a deterministic hash we need a standardised way of stringifying to a minimal relevantdocument.
-In doing so we also avoid producing different hashes for documents that contain a possible extra redundant character.
+To produce a deterministic `documentId` we need a standardised way of stringifying to a minimal relevantdocument.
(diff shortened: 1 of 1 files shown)


rfcs/PersistedOperations.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Reduce request transfer size (GraphQL documents have a deterministic size)
+- Support caching (avoid running into max-url size constraints when using the `GET` HTTP method)
+- Lock down the amount of request permutations (the server can choose to only accept persisted operations it is aware of)
+
+## Flow
+
+### Producing the document-id
+
+To produce a deterministic `documentId` we need a standardised way of stringifying to a minimal relevantdocument.
```
benjie, line 17:
```suggestion
To produce a deterministic `documentId` we need a standardised way of stringifying to a minimal relevant document.
```

benjie (approved):

benjie:
RFC documents get merged easily without much in the way of review.

PR #257, opened 2023-08-21
Address the URL limit for GET operations

tazsingh:
Some CDNs (such as Cloudfront) do not support caching over POST. Thus, it's dramatically easier to hit the URL limit when sending variables over a query string.

I've added a paragraph to mention utilizing the POST body as an option for this or a HTTP header. I've personally used a `X-GQLVARS` header. Happy to include that naming here or leave it to the implementer on what to use

As I opened this document, my Grammarly extension corrected wording in some places. I've fixed them as the extension recommended, however I'm also happy to revert those changes and focus solely on the URL limit.

~~Note: I'm also creating this PR in order to sign the EasyCLA to participate in GraphQL Foundation meetings~~


benjie (approved):
Thanks Taz! In future please send grammar/editorial fixes in a separate PR so they can be reviewed independently.

PR #260, opened 2023-10-03
Add myself to the October meeting

JoviDeCroock:


benjie:
One of the big outcomes for me during GraphQLConf is that almost all of the attendees should have been using persisted operations (**NOT** APQ), and yet almost none of them are. Many cite a lack of documentation around how to do it, and others got confused by the confusingly named automatic persisted queries (APQ) which forgo many of the benefits of persisted operations, only keeping the bandwidth savings. Reviving this topic seems like a good idea.

If we get enough attendance, I'd also like to move that we advance the GraphQL-over-HTTP RFC to the next stage.

CC: @abernix @balshor @benjie @deinok @erikwittern @jaydenseric @michaelstaib @mike-marcacci @mmatsa @shane32 @sjparsons @spawnia @sungam3r @touchstone117 @enisdenjo @JoviDeCroock

PR #262, opened 2023-10-03
Add "Advancing the spec to stage 2" to October WG

benjie:
I'd like to review remaining issues and discuss with implementors how they feel about the current status, especially now that we have a year of experience with it. I, personally, think it's pretty much ready to advance.

cc @enisdenjo especially.


enisdenjo (approved):
Exciting stuff!

Issue #38, opened 2019-11-21
"Persisted queries" / "stored operations"


benjie:
The terms "persisted queries" and "stored operations" are typically used interchangably. The traditional name is "persisted queries", but the term "stored operations" acknowledges that this approach can be used with mutation and subscription operations in addition to queries.

There's lots of different implementations of this feature, often with different aims. Sometimes it's used to block all but a small allow-list of operations. Sometimes it's used to reduce network bandwidth for GraphQL requests. Sometimes it's used to improve HTTP cacheability (GET requests with short URLs).

Relevant links:

- https://www.apollographql.com/docs/apollo-server/performance/apq/
- https://www.graphile.org/postgraphile/production/#simple-query-whitelist-persisted-queries
- https://hotchocolate.io/docs/persisted-queries
- https://github.com/restorando/graphql-query-whitelist

Some implementations require queries to be persisted to the server in advance of receiving queries (e.g. at build time); others negotiate the storage of the query with the server for bandwidth optimisation (e.g. automatic persisted queries).

Typically persisted queries replace the `query` in `{query, variables, operationName}` with another field (typically `id`) that indicates an identifier for the operation - perhaps a number or hash.

One of the optimisations that stored operations enable is that the `parse` and `validate` steps of GraphQL execution can be performed once per stored operation and cached; future requests for the same stored operation can jump straight to the `execute` step.

benjie:
@n1ru4l points out a more correct term is actually "stored _documents_"

n1ru4l:
We are currently building GraphQL Yoga v3 plugins for [Automatic persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/) and also traditional built-time extracted persisted documents.

I see Automatic persisted queries as an upstream client-to-origin payload size optimization and persisted documents as a security feature for preventing the processing of arbitrary GraphQL documents.

It seems like all urql and apollo-client have built-in support for Automatic persisted queries (which itself does also not have a real specification) and the "extensions-protocol" became the de-facto standard for both APQ and persisted document implementations (see mecurius).

However, we would rather prefer to use something for persisted operations that actually follows a specification. IMHO the GraphQL over HTTP specification would be perfect for this.

In https://github.com/enisdenjo/graphql-ws/issues/188#issuecomment-851847340 there has been a hint that this could simply be solved by introducing an `id` field that if set replaces the `query` field.

Shane32:
[...]

PR #264, opened 2023-10-09
[RFC] Add Appendix A: Persisted Documents

benjie:
Addresses #38 (though not the "automatic" version; though that can be built on top of this spec).

This is a first draft.


spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,256 @@
...
+For a _persisted document request_ using HTTP GET, parameters SHOULD be provided
+in the query component of the request URL, encoded in the
+`application/x-www-form-urlencoded` format as specified by the
+[WhatWG URLSearchParams class](https://url.spec.whatwg.org/#interface-urlsearchparams).
+
+The {documentId} parameter must be a string _document identifier_.
+
+The {operationName} parameter, if present, must be a string.
+
+Each of the {variables} and {extensions} parameters, if used, MUST be encoded as
```
JoviDeCroock, line 150: Do we have to define expectations in case this exceeds maximum URL size?

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,256 @@
...
+
+When using persisted documents as an operation allowlist, the client MUST
+persist the documents it uses ahead of time in a secure manner (preventing
+untrusted third parties from adding their own persisted document) such that the
+server will be able to retrieve the identified document within a _persisted
+document request_ and know that it is trusted.
+
+## Persisted Document Request
+
+A server MAY accept a _persisted document request_ via `GET` or `POST`.
```
JoviDeCroock, line 124: Do we define an opinion on mutations have to be POST to help folks avoid the pitfall of caching a mutation by accident

JoviDeCroock (approved):
I am super excited to see this come to fruition! Should we remove the RFC?

spec/Appendix A -- Persisted Documents.md
benjie: If we do; we should do it in the main spec: https://graphql.github.io/graphql-over-http/draft/#sec-GET

The Appendix tries not to redundantly repeat statements from the main spec if it can avoid it.

benjie: Which reminds me; I heard that some people are using headers to specify variables when using GraphQL-over-GET... Apparently that works around the length limit :raised_eyebrow:

spec/Appendix A -- Persisted Documents.md
benjie: You know; I figured this was already stated in the main spec so didn't need to be repeated... But you're right - it should be repeated. I'll add it.

spec/Appendix A -- Persisted Documents.md
JoviDeCroock: Yes, a lot of folks use headers and then the server adds them to the `Vary` response-headers so browsers/... can know that it is part of the cache-key

benjie:
Added to October 26th agenda: https://github.com/graphql/graphql-over-http/blob/main/working-group/agendas/2023/2023-10-26.md

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+## Persisted Document Response
+
+When a server that implements _persisted documents_ receives a well-formed
+_persisted document request_, it must return a well‐formed _GraphQL response_.
+
+The server should retrieve the GraphQL Document identified by the {documentId}
+parameter. If the server fails to retrieve the document, it MUST respond with a
+well-formed _GraphQL response_ consisting of a single error. Otherwise, it
```
Shane32, line 259: I suggest we define the contents of the error sufficiently for the client to conclusively recognize.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,262 @@
...
+## Persisted Document Request
+
+A server MAY accept a _persisted document request_ via `GET` or `POST`.
+
+### Persisted Document Request Parameters
+
+:: A _persisted document request_ is an HTTP request that encodes the following
+parameters in one of the manners described in this specification:
+
+- {documentId} - (_Required_, string): The string identifier for the Document.
```
Shane32, line 131: `query` is omitted here and throughout.  Makes sense, as it is not necessary for operation and should not be included.  But I think APQ-type behavior should be considered and allowed for within the spec.  Preferably, it is an optional feature as part of this spec.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,262 @@
...
+- {operationName} - (_Optional_, string): The name of the Operation in the
+  identified Document to execute.
```
martinbonnin, line 133: Is that even needed? If we're saying there should be a single operation in a persisted document, there's no ambiguity on which one to execute. Does the client need to send the `operationName`?

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,262 @@
...
+query($id:ID!){user(id:$id){name}}
+```
+
+Would have this different _SHA256 hex document identifier_:
+
+```example
+sha256:71f7dc5758652baac68e4a10c50be732b741c892ade2883a99358f52b555286b
+```
+
+### Custom Document Identifier
```
mohsen1, line 88: At Airbnb we append the operation name to the URL for the GraphQL requests for easier debugging and tracing. I wonder if this spec can include an example or implementation of such solutions:

Screenshot 2023-10-10 at 11 10 30 AM

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,262 @@
...
+implementation specific.
+
+Note: A 32 character hexadecimal _custom document identifier_ is likely to be an
+MD5 hash of the GraphQL document, as traditionally used by Relay.
+
+## Persisting a Document
+
+A client utilizing persisted documents MUST generate a _document identifier_ for
+each GraphQL Document it wishes to issue to the server, and SHOULD ensure that
+the server can retrieve this GraphQL Document given the document identifier.
```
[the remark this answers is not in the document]
benjie: I've rewritten this entire section.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,262 @@
...
+### Prefixed Document Identifier
+
+:: A _prefixed document identifier_ is a document identifier that contains at
+least one colon symbol (`:`). The text before the first colon symbol is called
+the {prefix}, and the text after it is called the {payload}. The {prefix}
+identifies the method of identification used. Applications may use their own
+identification methods by ensuring that the prefix starts `x-`; otherwise, all
+prefixes are reserved for reasons of future expansion.
+
+### SHA256 Hex Document Identifier
```
dotansimha, line 44: I wonder if actual encryption is needed? what's the benefit of having `sha256` over any other opaque string? 🤔 
We can allow users how to encode/map their persisted documents, and then do matching based on that value?

spec/Appendix A -- Persisted Documents.md
benjie: There's not much scope to do that currently; the best that we can do is ensure that the error message starts with, ends with, or contains, a particular string. Error codes would need to be specified in the main GraphQL specification for us to use them here, and writing to `extensions` from an official specification has all the issues previously raised. Since this is only needed for APQ currently, I'm happy leaving it unspecified (and addressing it when APQ is added), but I would support the addition of a non-normative example of the error message.

spec/Appendix A -- Persisted Documents.md
benjie: I think it would be unwise to recommend that people do this based on the operation name (the risk of clashes as they iterate their queries is too high); however I would support the name being factored into the document idea along with some hashing; e.g. something like: `createHash('sha256').update(documentSource).digest('hex').substring(0, 12) + '_' + operationName`. The `operationName` query parameter is already specified; perhaps we should add a non-normative note recommending that clients include it to aid in debugging?

spec/Appendix A -- Persisted Documents.md
benjie: The aim of a specification like this is interoperability; so a client that supports persisted operations should be able to use a server that supports persisted operations without too much additional setup. Sharing details of the document identification method used out-of-band is supported (explicitly by `x-` prefix, or by custom identifiers); but for maximum compatibility there should be a shared baseline in my opinion.

benjie:
Thanks for your feedback everyone! I've adopted some of the feedback and replied to others. Keep it coming!

mcollina (approved):
lgtm, this is amazing

spec/Appendix A -- Persisted Documents.md
Shane32: I expect that the benefit is mainly for APQ support, where a client can enable APQ without knowledge of the server implementation, and so a consistent implementation is essential.  SHA256 can be computed natively by modern browsers with a very low collision rate, making it a good choice in this scenario.

However, in a scenario where the identifiers are only known to the server, and must be registered with the server in order for the client to operate (as so far is documented in this RFC), they might as well be any opaque key returned by the query storage database.  If the database stores the queries with an auto-incrementing integer as an identifier, that would work just as well.

Even so, there are still some benefits to using a hash:
1. Should the query storage mechanism change, the identifiers will remain constant (but the same would be true of a GUID)
2. Inherit deduplication for stored queries
3. Enhanced security, as attackers can't easily guess valid identifiers, reducing the risk of unauthorized query execution

I'm fine with the current suggestion of `sha256:` but allowing for `x-id:` and similar.

benjie:
I've also written a piece on "trusted documents" - essentially the type of persisted documents that you can build an allowlist with (also known as "persisted queries" and various other names):

https://benjie.dev/graphql/trusted-documents

My hope is that "trusted documents" can become the preferred term when persisted documents are used as an allowlist (i.e. where your developers have written them), since it cannot be confused with other techniques such as "automatic persisted queries" (APQ), since these techniques involve no trust.

spec/Appendix A -- Persisted Documents.md
dotansimha: I see, thank you @Shane32 @benjie .
My point is not about the prefix or the method but for the need for encryption? a user can decide to use `operation-1` as the key (instead of an actual computed hash), and the result will be the same 🤔 🤔 🤔

benjie: @dotansimha Indeed, that's already allowed under this spec. The issue is it requires coordination between server and client (they need to agree on how `operation-1` is derived). For maximum compatibility, if both server and client already know how to identify operations (e.g. standardized SHA256 hash) then all that's left is to transfer the docs from client to server, which can happen _after_ the client has been built, rather than before or during, and no configuration is required on server or client easing adoption for everyone and encouraging more people to use an operation allowlist.

benjie: > However, in a scenario where the identifiers are only known to the server, and must be registered with the server in order for the client to operate (as so far is documented in this RFC)

I don't think that's what the spec says...

> A service which accepts a _persisted document request_ SHOULD support the
_SHA256 hex document identifier_ for compatibility.

^ So a client SHOULD be able to default to using the SHA256 hash, and will only need to be customized if that's not the case.

dotansimha (approved):

spec/Appendix A -- Persisted Documents.md
benjie: Oh, right, but the queries still need to be transferred to the server is what you're saying. Yes, that's true, but it can be done _after_ the client is built (but before it's deployed). That's different from having to do it _during_ the build process; it means that clients can be built and persisted documents written even before the server _exists_. It also allows for arbitrary transfer of documents to the server (you can send them one-way on a pen drive through the mail if you want!).

martinbonnin:
Regarding `documentId`, would it make sense to add it to the main spec spec page as well? Somewhere around [here](https://github.com/graphql/graphql-over-http/blob/f1f7ea60cd037f396205e6ab2907251d41c6d961/spec/GraphQLOverHTTP.md?plain=1#L208), in the "Request parameters" section?

If `documentId` is a reserved key, might as well allocated it in the main spec page? Going to an appendix for that feels a bit off.

benjie:
@martinbonnin I see what you mean, but `documentId` is meaningless to a _GraphQL-over-HTTP request_; it's only relevant to a _persisted document request_. I guess we could add it as a "reserved key" ("a _GraphQL-over-HTTP request_ must never contain a property `documentId`"); but at the moment _all_ unspecified keys are effectively reserved (if you need to add a key, it should go in `extensions`) so that seems redundant.

martinbonnin:
@benjie right. Maybe "Appendix B: Reserverd keys" 🙃? I don't really know... 
It'd be interesting to compare to what other protocols with optional features have done in the past. I don't have an example at hand right now but I'll keep looking.

spec/Appendix A -- Persisted Documents.md
benjie: "Recommend" and "should" are the same according to RFC2119, so I think that is saying what's already there. If the server follows this recommendation, the client doesn't need any configuration. If the server _doesn't_ do this; _then_ you need to coordinate between client and server. For optimal interoperability, no coordination should be necessary.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+
+This appendix defines an optional extension to the GraphQL-over-HTTP protocol
+that allows for the usage of "persisted documents".
+
+:: A _persisted document_ is a GraphQL document (strictly: an
+[`ExecutableDocument`](https://spec.graphql.org/draft/#ExecutableDocument)) that
+has been persisted such that the server may retrieve it based on an identifier
+indicated in the HTTP request.
+
+This feature can be used as an operation allow-list, as a way of improving the
```
JoviDeCroock, line 11: Should we go for enabling, I know that HTTP Get and URL encoding the query-string enables the same thing but it's limiting while this just blanket enables you to do this

```suggestion
This feature can be used as an operation allow-list, as a way of enabling the
```

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+Note: When used as an operation allowlist, persisted documents are typically
+stored into some kind of trusted shared key-value store at client build time
+(either directly, or indirectly via an authenticated request to the server) such
+that the server may retrieve them given the identifier at request time. This
+must be done in a secure manner (preventing untrusted third parties from adding
+their own persisted document) such that the server will be able to retrieve the
+identified document within a _persisted document request_ and know that it is
+trusted.
+
+Note: When used solely as a bandwidth optimization, an error-based mechanism
```
JoviDeCroock, line 115: As we hint here at APQ should we have an opinion on the runtime vs build time generation of documents where we explicitly state that build-time gets you all the security benefits while runtime does not?

spec/Appendix A -- Persisted Documents.md
benjie: Out of an abundance of caution I think we should avoid this change to prevent it being used as evidence that "GraphQL doesn't do caching" without X, Y and Z. It's a nice edit though. Perhaps we could soften it, `as a simple way of setting up caching of GraphQL operations` or something...

spec/Appendix A -- Persisted Documents.md
benjie: I'm happy with the wording as-is to cover this. Note that Persisted Documents are not necessarily _trusted_ documents. All trusted documents are persisted documents, but not all persisted documents are trusted documents.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+client issues a follow-up request containing the full GraphQL Document to be
+persisted.
+
+Note: When persisting a document it is generally good practice for the client to
+issue both the GraphQL Document and the document identifier to the server; the
+server would then regenerate the document identifier from the GraphQL Document
+independently, and check that the identifiers match before storing the Document.
+An alternative but equally valid approach has the client issue the GraphQL
+Document to the server, and the server returns an arbitrary _custom document
+identifier_ that the client would incorporate into its bundle.
```
n1ru4l, line 127: We should write here that the operation identifier to persisted document mapping is usually a build artifact shared by the graphql client and graphql server.

The server can use an external "store" as the source for resolving a document identifier to an actual document

spec/Appendix A -- Persisted Documents.md
n1ru4l: What are your opinions on providing the `documentId` via the URL `/graphql/<documentId>`?
One of the benefits would be easier debugging and visibility in dev tooling.

spec/Appendix A -- Persisted Documents.md
benjie: Would you like to propose edits in the form of a PR? Perhaps you're just suggesting an extension to this paragraph, such as:

> identifier_ that the client would incorporate into its bundle. Either way, the operation identifier and persisted document mappings are usually a build artefact shared by the client and server.
>
> Note: The server will typically retrieve persisted documents from a "store" at run-time, using the identifier as the lookup. The store could be files on the filesystem, a database, a durable in-memory key-value store, or anywhere else suitable for retrieving a value by a key.

spec/Appendix A -- Persisted Documents.md
benjie: Yes I think this should be encouraged more; it's great for caching. I'd welcome your edits to address this, if you were so inclined.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+colon symbols (`:`). The meaning of a custom document identifier is
+implementation specific.
+
+Note: A 32 character hexadecimal _custom document identifier_ is likely to be an
+MD5 hash of the GraphQL document, as traditionally used by Relay.
+
+## Persisting a Document
+
+A client that wishes to utilize persisted documents for a request must generate
+a _document identifier_ for the associated GraphQL Document and should ensure
```
martinbonnin, line 101: Slightly itched by the "generate" wording here. In most scenarios I've seen, the server actually generates the id. Can we rework that paragraph to make it a bit more "symmetrical"? Not really sure how but something like this maybe?

```
In order to use persisted documents, the client and server must share the _document identifier_ to _persisted document_ association so that a server can retrieve this GraphQL Document from the document identifier. The
method through which the client and server achieve this is implementation
specific.
```

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+the server can retrieve this GraphQL Document from the document identifier. The
+method through which the client and server achieve this is implementation
+specific.
+
+Note: When used as an operation allowlist, persisted documents are typically
```
martinbonnin, line 106: Nit for consistency with the previous wording (or just allowlist everywhere?)

```suggestion
Note: When used as an operation allow-list, persisted documents are typically
```

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+stored into some kind of trusted shared key-value store at client build time
```
martinbonnin, line 107: Nit: 

```suggestion
stored into a trusted shared key-value store at client build time
```

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+(either directly, or indirectly via an authenticated request to the server) such
```
martinbonnin, line 108: > either directly, or indirectly via an authenticated request to the server

Not sure to understand the difference? Wouldn't both method always require some sort of  authenticated request?

spec/Appendix A -- Persisted Documents.md
martinbonnin: Should we mention "auto persisted queries" explicitely? 

```
Note: When used solely as a bandwidth optimization, sometimes referred as "auto persisted queries", ...
```

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+Note: When used solely as a bandwidth optimization, an error-based mechanism
+might be used wherein the client assumes that the document has already been
+persisted, but if the request fails due to unknown _document identifier_ the
+client issues a follow-up request containing the full GraphQL Document to be
+persisted.
+
+Note: When persisting a document it is generally good practice for the client to
+issue both the GraphQL Document and the document identifier to the server; the
+server would then regenerate the document identifier from the GraphQL Document
+independently, and check that the identifiers match before storing the Document.
```
martinbonnin, line 124: Should we say "good practice" here? IMO, where the algorithm to derive an identifier sits is completely symmetrical?

1. on the client => client sends both document + id
2. on the server => client sends document, retrieves id
3. both => both client and server share the same algorithm (sha256)

spec/Appendix A -- Persisted Documents.md
martinbonnin, line 127: >  a build artifact shared by the graphql client and graphql server.

In case of multiple clients, a given client only has a "partial" view of the store so it's technically not the same artifact. 

> an arbitrary custom document identifier that the client would incorporate into its bundle.

`bundle` is a very web-centric term. Maybe just:

```
an arbitrary custom document identifier that is included in the client.
```

?

So that it works equally for mobile/desktop/cli clients?

spec/Appendix A -- Persisted Documents.md
martinbonnin: +1 to non-normative example. Attempt:

```
Note: typically, the error allows recognizing failures to retrieve a document:

\```json
{
  "errors": [{
    "message: "unknown document identifier"
  }]
}
\```

Implementation may add a dedicated error code in the response extensions as described in [the error result format](https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format)
```

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Persisted Document Response
+
+When a server that implements _persisted documents_ receives a well-formed
+_persisted document request_, it must return a well‐formed _GraphQL response_.
+
+The server should retrieve the GraphQL Document identified by the {documentId}
+parameter. If the server fails to retrieve the document, it MUST respond with a
+well-formed _GraphQL response_ consisting of a single error. Otherwise, it
+should construct a _GraphQL-over-HTTP request_ using this document and the other
```
martinbonnin, line 267: Should the "should" be a must?

spec/Appendix A -- Persisted Documents.md
martinbonnin: To be pedantic there, I'd argue that sharing the "sha256" method is still coordination between the client and server. Plus the documents need to be actually transferred (on a pen drive or avian carrier!). 

So coordination is always required but sha256 is a convenient and widespread default which we recommend (hence the should)?

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,269 @@
...
+:: A _document identifier_ is a string-based identifier that uniquely identifies
+a GraphQL Document.
+
+Note: A _document identifier_ must be unique, otherwise there is a risk of
+responses confusing the client. Even if the selection sets are identical, even
+whitespace changes may change the location from which errors are raised, and
+thus should generate different document identifiers.
+
+A _document identifier_ must either be a _prefixed document identifier_ or a
+_custom document identifier_.
```
martinbonnin, line 33: Should we give a formal BNF syntax? Maybe restrict the identifiers to alpha numeric? GraphQL names maybe?

spec/Appendix A -- Persisted Documents.md
benjie: Indeed, it is a very light-touch asynchronous form of coordination. Essentially the coordination boils down to two things: 1. does the server support SHA256 hashes (don't necessarily need to ask the server this, it's a fact that should be established in the development team); 2. we need a way to send the operations+hashes to the server and to know when they have been persisted.

Client is informed that server supports SHA256 hashes.
Client: _performs build including generation of hashes_
Client (at some point later): _ships hashes and operations to server somehow_
Server (upon receipt): _stores queries+hashes somehow_
Client (after server has stored): _is deployed_

Since the client can't be deployed until the server has stored the queries/hashes, there is indeed coordination. The coordination is incredibly lightweight compared to alternatives where the server must generate hashes during the client build process.

spec/Appendix A -- Persisted Documents.md
benjie: I've reworded the paragraph to:

```md
To utilize persisted documents for a request, the client must possess a unique
_document identifier_ for the associated GraphQL Document, and the server must
be able to retrieve this GraphQL Document using the document identifier. The
method through which the client and server achieve this is implementation
specific.
```

spec/Appendix A -- Persisted Documents.md
benjie: 1. On the client via custom algorithm: client must send document and ID.
2. On the server: client cannot send ID since it doesn't know it yet.
3. On the client via standard SHA256 algorithm: client SHOULD send document and ID, but technically it only needs to send the document since the server can regenerate the ID. Better if the client sends both, then the server can validate it's SHA256 hash matches the client's one, detecting issues (or bad actors) earlier.

I see your point though, so I'll reword the note:

```md
Note: When persisting a document for which the identifier has been derived by
the client, it is generally good practice for the client to issue both the
GraphQL Document and the document identifier to the server; the server could
then regenerate the document identifier from the GraphQL Document independently,
and check that the identifiers match before storing the Document. If the
identifier is not derived on the client then the client must coordinate
retrieval of a document identifier from the server to be incorporated into its
bundle.
```

spec/Appendix A -- Persisted Documents.md
benjie: I don't see the term "build artifact" in the RFC?

Acknowledged on the second point; incorporated into above edit.

spec/Appendix A -- Persisted Documents.md
benjie: No; it can refuse to do so on security grounds (or any other grounds). For example, if the given identifier were `../../../../../../etc/passwd` :wink:

benjie: Sorry... that comment was on the wrong `should`. This one... Let me think.

benjie: No, I think `should` is still fine. The server might have a reason to not perform the request, for example if the given document identifier is only permitted to be used by admins, and the request doesn't come from an admin.

Though I suppose you could cover that under "fails to retrieve the document" due to scoping issues...

Sure; let's set it to MUST.

benjie: I've avoided it entirely:

```
well-formed _GraphQL response_ consisting of a single error. Otherwise, it will
construct a _GraphQL-over-HTTP request_ using this document and the other
parameters of the _persisted document request_, and then follow the details in
the [Response section](#sec-Response).
```

martinbonnin: Makes sense 👍

spec/Appendix A -- Persisted Documents.md
martinbonnin: > I don't see the term "build artifact" in the RFC?

It was a response to [this comment](https://github.com/graphql/graphql-over-http/pull/264#discussion_r1608533030). Sorry it's out of context but the spec is all good 👍 

> Acknowledged on the second point; incorporated into above edit.

Thanks!

spec/Appendix A -- Persisted Documents.md
benjie: Would you like to submit a PR to my PR to add this?

martinbonnin: Sounds like a plan. This week's quite busy but I'll aim for next week! (famous last words 😅 )

martinbonnin: That was the longest week ever but attempt at formal syntax is [here](https://github.com/graphql/graphql-over-http/pull/296/files)

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,271 @@
...
+Setting the value of the {operationName} parameter to the empty string is
+equivalent to omitting the {operationName} parameter.
+
+A client MAY provide the _persisted document request_ parameters in another way
+if the server supports that.
+
+Note: A common alternative pattern is to use a dedicated URL for each _persisted
+operation_ (e.g.
+`https://example.com/graphql/sha256:71f7dc5758652baac68e4a10c50be732b741c892ade2883a99358f52b555286b`).
+
```
JoviDeCroock, line 171: Can we add a section on i.e. returning 404 for persisted-documents we can't find and maybe even 400 if they don't leverage the right prefix?

benjie: I think for the URL format 404 should be encouraged. I've been writing up a change to the appendix which encourages the URL format, but haven't had time to finish it yet; I've just raised a PR for my WIP so we have something to easily reference: https://github.com/graphql/graphql-over-http/pull/305

IMO for the non-URL version (traditional), 404 should not be used - it suggests that the `/graphql` endpoint is not found, which would be confusing.

spec/Appendix A -- Persisted Documents.md
ddebrunner: Was coming to ask if there was a consistent error when the document identifier is not found

Shane32:
This RFC for Persisted Documents as currently written is now supported in [GraphQL.NET 8.0.1](https://www.nuget.org/packages/GraphQL) and [GraphQL.NET Server 8.0.1](https://www.nuget.org/packages/GraphQL.Server.Transports.AspNetCore).

Notes:
- Supported for GET, JSON POST, multipart POST (when enabled), and websocket requests (both [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) and [graphql-ws](https://github.com/enisdenjo/graphql-ws))
- No automatic persisted query support has been added for this spec
- Apollo Automatic Persisted Queries are still supported using the old spec (within `extensions`)
- Does not support url path parsing such as demonstrated in #305 yet
- Does not restrict character sets for document identifiers (see #296 )

spec/Appendix A -- Persisted Documents.md
```diff
@@ -0,0 +1,271 @@
...
+GET requests MUST NOT be used for executing mutation operations. If a mutation
```
norbertosus, line 172:
```suggestion
GET requests MUST NOT be used for executing mutation operations. If a mutation
```

ghost (approved):

ghost (changes requested):
>

spec/GraphQLOverHTTP.md
```diff
@@ -738,3 +738,5 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
 well-formed _GraphQL response_ since the source of the response may not be the
 server but instead some intermediary such as API gateways, proxies, firewalls,
 etc.
+
+# [Appendix: Persisted Document](Appendix%20A%20--%20Persisted%20Documents.md)
```
glen-84:
```suggestion
# [Appendix: Persisted Documents](Appendix%20A%20--%20Persisted%20Documents.md)
```

PR #268, opened 2023-10-26
Update Zoom link in today's agenda

benjie:
No idea why the old one expired.


benjie:
cc @enisdenjo @ghmcadams

PR #278, opened 2023-10-27
Make it clear that other keys are reserved

### LINKED ISSUE #271: Make it clear that extra keys in the request/response payloads are not allowed
This is a task for myself really.

The `extensions` object exists so that vendors may add their own properties to requests/responses/errors/etc. All top-level keys are reserved by the spec. We should make this explicit, such that adding `id` as a key to a request automatically makes the request _not_ a GraphQL-over-HTTP request. (Which is fine, because the server is still allowed to handle it however it sees fit.)

benjie:
Fixes #271

The `extensions` object exists so that vendors may add their own properties to requests/responses/errors/etc. All top-level keys are reserved by the spec. This PR makes this explicit for JSON-encoded requests.

@enisdenjo Will this break any of the existing implementations, other than Relay's persisted queries network layer recommendation?

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -376,6 +376,11 @@ When encoded in JSON, a _GraphQL-over-HTTP request_ is encoded as a JSON object
   names and the values of which are the variable values
 - {extensions} - an optional object (map)
 
+All other property names are reserved for future expansion; if implementors need
+to add additional information to a request they MUST do so via other means, the
+RECOMMENDED approach is to add an implementor-scoped entry to the {extensions}
+object.
+
 ### Example
 
 If we wanted to execute the following GraphQL query:


Shane32 (approved):

martinbonnin (approved):

JoviDeCroock (approved):

enisdenjo:
I don't think this will break anything. Looks good to me!

It is kinda conventionalized that all metadata lands on the `extensions` field.

enisdenjo (approved):

enisdenjo:
Follow-up question: should I add a test to graphql-http reflecting this? It's now illegal to have extraneous keys in the request body.

benjie:
Servers don't need to validate this, and in fact they should not since a client from a future spec version may send additional keys, which legacy servers should just ignore.

enisdenjo:
Hmm, is the "must" then confusing?

> [...] if implementors need to add additional information to a request they MUST do so via other means,

If we add more fields in future spec versions, we can just add the fields/relax the compliance tests?

Shane32:
> Servers don't need to validate this, and in fact they should not since a client from a future spec version may send additional keys, which legacy servers should just ignore.

I think we should add this to the spec.  The main graphql spec did not state this explicitly for reserved names (those starting with `__`), and now any current GraphQL UI applications built on Apollo's graphql library crash if they see additional reserved names in the introspection result.  (I consider this a bug but 🤷‍♂️ ).  They should be built to ignore additional features added to the spec, and the spec should state this.

We could add something like this:

_Servers receiving a request with additional properties MUST ignore the additional properties when processing the request._

benjie:
> and now any current GraphQL UI applications built on Apollo's graphql library crash if they see additional reserved names in the introspection result. (I consider this a bug but 🤷‍♂️ )

That must be a bug. All potential field names that start with `__` are reserved for introspection usage, it must be anticipated that we can use additional such fields in future.

PR #290, opened 2024-03-27
GraphQLOverSSE: note including an empty data field in the complete message

enisdenjo:
https://github.com/enisdenjo/graphql-sse/pull/71

### DIFF
--- a/rfcs/GraphQLOverSSE.md
+++ b/rfcs/GraphQLOverSSE.md
@@ -48,6 +48,9 @@ interface CompleteMessage {
 }
 ```
 
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


bclozel:
Is there a consistent way to detect whether clients use `EventSource`, or is this really a local implementation choice? If there is no consistent way, shouldn't the server always send `data: ` with the complete event or are there downsides to doing that?

enisdenjo:
@bclozel there is no way to detect the client. You could manually indicate to the server through headers, but EventSource in browser doesnt allow changing headers.

graphql-sse always sends the `data: `, there is no reason against it. The only drawback is you send 6 more bytes, but I dont think this is a biggie.

Issue #48, opened 2019-12-17
X-GraphQL-Event-Stream for being notified that the GraphQL schema has updated


benjie:
[Server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) are a perfect mechanism for informing a GraphQL client that the schema has been updated. They don't require the complexity of websockets, and they are uni-directional.

I'm proposing that when introspecting a GraphQL API, if the header `X-GraphQL-Event-Stream` is detected then interested clients should subscribe to the `text/event-stream` at that (relative or absolute) URL and when it receives the `change` event it should automatically re-introspect the GraphQL schema.

Not much code should be required to implement this, just something like:

```js
// refreshSchema is a function you provide which triggers re-introspection

const streamUrl = response.headers["x-graphql-event-stream"];
if (streamUrl) {
  const endpointUrl = new URL(endpoint);
  const streamUrl = new URL(streamUrl, endpointUrl);
  if (endpointUrl.host !== streamUrl.host) {
    throw new Error(
      `Stream and endpoint hosts don't match - '${streamUrl.host}' !== '${endpointUrl.host}'`
    );
  }
  const eventSource = new EventSource(streamUrl);

  eventSource.addEventListener("change", refreshSchema, false);
  eventSource.addEventListener("open", () => { /* ... */ }, false);
  eventSource.addEventListener("error", () => { /* ... */ }, false);
}
```

(You may want to throttle the refreshSchema method.)

This has been baked into PostGraphile for a while; I opened [an implementation PR](https://github.com/skevy/graphiql-app/pull/120) to @skevy's GraphiQL.app a good while ago; and @imolorhe has [implemented it independently](https://github.com/imolorhe/altair/commit/0b3ccc188597efbe1a57e8f026071743ff6ebffc) in Altair. Samuel wrote more about it here: https://sirmuel.design/a-better-graphql-developer-experience-with-x-graphql-event-stream-1256aef96f24. We'll probably be implementing it in the new version of GraphiQL too, once we have the plugin system in place.

What are your thoughts?

sungam3r:
sometimes this feature can come in handy

imolorhe:
@benjie do you have a simple GraphQL server that has this implemented for demo purposes? I've been finding it hard to see this implementation actually working

benjie:
No; but it is implemented in a separate small file:

[...]

PR #296, opened 2024-07-21
Add identifier syntax

martinbonnin:
See https://github.com/graphql/graphql-over-http/pull/264#discussion_r1615049559

### DIFF
--- a/spec/Appendix A -- Persisted Documents.md
+++ b/spec/Appendix A -- Persisted Documents.md
@@ -95,6 +95,35 @@ implementation specific.
 Note: A 32 character hexadecimal _custom document identifier_ is likely to be an
 MD5 hash of the GraphQL document, as traditionally used by Relay.
 
+### Document identifier syntax
+
+DocumentIdentifier ::
+
+- PrefixedDocumentIdentifier
+- CustomDocumentIdentifier
+
+PrefixedDocumentIdentifier ::
+
+- Sha256HexDocumentIdentifier
+- OtherPrefixedDocumentIdentifier
+
+Sha256HexDocumentIdentifier :: sha256 `:` Sha256Checksum
+
+Sha256Checksum :: 64 LowerCaseHexDigit
+
+LowerCaseHexDigit :: one of
+
+- `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`
+- `a` `b` `c` `d` `e` `f`
+
+OtherPrefixedDocumentIdentifier :: x- IdentifierCharacter+ `:` IdentifierCharacter+
+
+CustomDocumentIdentifier :: SourceCharacter+
+
+IdentifierCharacter :: SourceCharacter but not `:`
+
+SourceCharacter :: "Any Unicode scalar value"
+
 ## Persisting a Document
 
 To utilize persisted documents for a request, the client must possess a unique


spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,35 @@ implementation specific.
...
+
+- PrefixedDocumentIdentifier
+- CustomDocumentIdentifier
+
+PrefixedDocumentIdentifier ::
+
+- Sha256HexDocumentIdentifier
+- OtherPrefixedDocumentIdentifier
+
+Sha256HexDocumentIdentifier :: sha256 `:` Sha256Checksum
```
Shane32, line 110:
```suggestion
Sha256HexDocumentIdentifier : `sha256:` Sha256Checksum
```

spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,35 @@ implementation specific.
...
+
+Sha256Checksum :: 64 LowerCaseHexDigit
+
+LowerCaseHexDigit :: one of
+
+- `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`
+- `a` `b` `c` `d` `e` `f`
+
+OtherPrefixedDocumentIdentifier :: x- IdentifierCharacter+ `:` IdentifierCharacter+
```
Shane32, line 119:
```suggestion
OtherPrefixedDocumentIdentifier :: 'x-' IdentifierCharacter+ `:` IdentifierCharacter+
```

spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,35 @@ implementation specific.
...
+
+CustomDocumentIdentifier :: SourceCharacter+
+
+IdentifierCharacter :: SourceCharacter but not `:`
+
+SourceCharacter :: "Any Unicode scalar value"
```
Shane32, line 125:
```suggestion
SourceCharacter :: Any Unicode scalar value
```

Shane32 (commented):
Perhaps only use quotes around strings and characters?

spec/Appendix A -- Persisted Documents.md
martinbonnin: Why not backticks here?

spec/Appendix A -- Persisted Documents.md
martinbonnin: I took over this one from the main spec ([source](https://github.com/graphql/graphql-spec/blob/3adfcca73644234fbbbb062a5cec9e7703419a9f/spec/Appendix%20B%20--%20Grammar%20Summary.md?plain=1#L5))

Overall, agree more consistency would be good here. Backticks for terms that are actually part of the syntax as-is and no quote/tick for sentences that are explicative but should not appear in the constructed element?

spec/Appendix A -- Persisted Documents.md
martinbonnin: I've pushed backticks in https://github.com/graphql/graphql-over-http/pull/296/commits/81fa52953cbafe0e3f9a8c41a39f707cd12b68fc

spec/Appendix A -- Persisted Documents.md
martinbonnin: I 've pushed backticks and no-quotes in https://github.com/graphql/graphql-over-http/pull/296/commits/81fa52953cbafe0e3f9a8c41a39f707cd12b68fc

benjie:
It's interesting you should raise this now; I was working on Friday on a reworking of the persisted documents spec that relates to using the `/graphql/<hash>/<operationId>` URLs and so I also felt this need to define the identifiers. I think we should encourage these URLs as the default going forward, but that's a topic for a different PR; I'm not ready to open that PR just yet but here's the commit where I added the syntax (note that I also narrowed the allowed characters to make them URL-safe):

https://github.com/graphql/graphql-over-http/commit/a3d9ecb9e988a22cb9010618becbc67117710ec8

I present this only for comparison, in case it's helpful.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,35 @@ implementation specific.
...
+- CustomDocumentIdentifier
+
+PrefixedDocumentIdentifier ::
+
+- Sha256HexDocumentIdentifier
+- OtherPrefixedDocumentIdentifier
+
+Sha256HexDocumentIdentifier : `sha256:` Sha256Checksum
+
+Sha256Checksum :: 64 LowerCaseHexDigit
```
benjie, line 112: Can you point to anything in spec-md or the GraphQL spec that uses this syntax to say "64 of this thing"? I'd expect something more like `LowerCaseHexDigit{64}` but I had a quick scan last week and didn't see anything about specific repetition counts. In the end I decided to got be as strict on the _syntax_ of identifiers and leave that for _validation_ - the errors you get out of validation can be a lot more helpful than the ones you get from syntax.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,35 @@ implementation specific.
...
+- `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`
+- `a` `b` `c` `d` `e` `f`
+
+OtherPrefixedDocumentIdentifier :: `x-` IdentifierCharacter+ `:` IdentifierCharacter+
+
+CustomDocumentIdentifier :: SourceCharacter+
+
+IdentifierCharacter :: SourceCharacter but not `:`
+
+SourceCharacter :: Any Unicode scalar value
```
benjie, line 125: This is incredibly broad; I know that's effectively what we have currently but I feel like we should lock it down a bit.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,35 @@ implementation specific.
...
+
+- PrefixedDocumentIdentifier
+- CustomDocumentIdentifier
+
+PrefixedDocumentIdentifier ::
+
+- Sha256HexDocumentIdentifier
+- OtherPrefixedDocumentIdentifier
+
+Sha256HexDocumentIdentifier : `sha256:` Sha256Checksum
```
benjie, line 110:
```suggestion
Sha256HexDocumentIdentifier :: `sha256:` Sha256Checksum
```

spec/Appendix A -- Persisted Documents.md
martinbonnin: > Can you point to anything in spec-md or the GraphQL spec that uses this syntax to say "64 of this thing"? 

This is completely made up. I don't think we have anything like that at the moment.

> In the end I decided to not be as strict on the syntax of identifiers and leave that for validation

Works for me 👍 , I'll amend the PR

martinbonnin: https://github.com/graphql/graphql-over-http/pull/296/commits/3fbc34c7dac239eaa6ec3778e13abfe34fd3e077

spec/Appendix A -- Persisted Documents.md
Shane32: While we could allow all Unicode characters, the main spec defines a specific set of characters for use in that document.  Something similar could be done here. 

[SourceCharacter](https://spec.graphql.org/October2021/#SourceCharacter)
U+0009
U+000A
U+000D
U+0020–U+FFFF

On a side note, the spec is defining these characters as UTF-16 codes, not Unicode code points. 

I’m fine with allowing any Unicode character in this context, or any non control character would be fine too.

martinbonnin: I had only alphanumericals initially and then backtracked to "any unicode" because "why not". But I think your point about having the identifier in url path is the important one. I'll move to alphanumerical + a bunch of unreserved chars like you did [here](https://github.com/graphql/graphql-over-http/commit/a3d9ecb9e988a22cb9010618becbc67117710ec8). Questions: 

* do we even need `~`, `.`, `_`, `-`?
* do we need uppercase letters? This might cause issues on case-insensitive filesystems (hello macOS 👋 )

Shane32: I strongly recommend a wide range of characters. URLs can escape any Unicode character so there need be no limit on what characters can be used for document ids. Also, url paths are by definition case sensitive; but they may be treated with no case sensitivity. For instance, if the document id was a guid it would need dashes. If it was base64 it would need lowercase and uppercase and a couple extra characters. If it was readable names, it would need to support Unicode for foreign languages. And a custom implementation may decide to treat the document id with no case sensitivity.

martinbonnin: > if the document id was a guid it would need dashes

Excellent point.

> If it was base64 it would need lowercase and uppercase and a couple extra characters

Another excellent point. `+`, `/` and `=` would be needed too. Which would make them a bit awkard to use as url path segments... Interesting!

> If it was readable names, it would need to support Unicode for foreign languages

I think I disagree on this one. I don't  really see anyone using readable names as identifiers? Sounds like a footgun to me that if we can, we should discourage. Just like the relay cursors are encouraged to be opaque, I would encourage the document ids to be similarly opaque.

Tangential: I have a GraphQL-over-HTTP meeting in my calendar on Thursday. Is that still happening? Want to discuss over there?

Shane32: Personally, the most I would exclude is control characters, colon, and the two slashes. Or otherwise similar to most OS filename conventions.

martinbonnin: My current mental model is:

* OS filenames are user interface. They are visible to the user and should be human readable. 
* POs identifiers are technical and should be machine readable. 

I don't really see myself using french diacritics in my POs ids (although that could make for a fun API!). But again all the POs I have ever used are sha256 so maybe there are other use cases out there.

benjie: > Another excellent point. +, / and = would be needed too. Which would make them a bit awkard to use as url path segments... Interesting!

There's standard URL-safe base64 encoding that uses `-` and `_`: https://datatracker.ietf.org/doc/html/rfc4648#section-5 (the padding can be omitted).

> I think I disagree on this one. I don't really see anyone using readable names as identifiers? Sounds like a footgun to me that if we can, we should discourage. 

Very strong agree that persisted document IDs should not be hand-crafted. They should be hashes or counters or similar - computer assigned.

>  I have a GraphQL-over-HTTP meeting in my calendar on Thursday. Is that still happening? Want to discuss over there?

I've just created agenda files for them so feel free to add yourself if you want :+1: 

> Personally, the most I would exclude is control characters, colon, and the two slashes. Or otherwise similar to most OS filename conventions.

I don't see much value in this. In general it's best to be strict first and loosen up when a need arises, since you cannot become stricter later without it being a breaking change. What use case do you have for being defined loosely? Our intended use cases are GUIDs/UUIDs, hashes and counters, and those are typically numeric, hexadecimal, or base64 encoded. I can imagine namespacing them too (e.g. `MyApp.f9db3fbb3cfe...`) but other than the a  I don't think they should be much wider than a GraphQL `Name`.

martinbonnin: 👍 I added my name [for Thursday's meeting](https://github.com/graphql/graphql-over-http/pull/298)

Shane32: > What use case do you have for being defined loosely?

Personally, none at all. Just nowhere else are there similar restrictions on identifiers. Filenames have a very few restrictions. And here we can’t use colon or the string would be confused with a prefixed identifier. So I would wonder why we are further restricting these identifiers which are otherwise opaque to the application?

Comparing to GraphQL names, there it’s part of a formatted document where a parser must be able to read the identifier separate from other tokens. But here all these identifiers are already escaped by JSON just like any other string. As part of a url path, again there is a well defined encoding scheme available. 

Personally, I’d probably use url-safe base64 and period or fewer characters.

spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,50 @@ implementation specific.
...
+CustomDocumentIdentifier :: UnreservedCharacter+
+
+UnreservedCharacter ::
+
+- Letter
+- Digit
+- `-`
+- `.`
+- `_`
+- `~`
```
Shane32, line 129: Note: these 'Unreserved Characters' mirror the list within RFC 3986 here:

https://datatracker.ietf.org/doc/html/rfc3986#section-2.3

ALPHA and DIGIT are described further here:

https://datatracker.ietf.org/doc/html/rfc2234#section-6.1

Shane32: This encoding allows for GUIDs encoded as described by RFC 4122:

https://datatracker.ietf.org/doc/html/rfc4122#section-3

And it allows for base64url encoding as described by RFC 4648: (assuming no padding)

https://datatracker.ietf.org/doc/html/rfc4648#section-5

spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,50 @@ implementation specific.
...
+
+Sha256HexDocumentIdentifier :: `sha256:` LowerCaseHexDigit+
+
+LowerCaseHexDigit :: one of
+
+- `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`
+- `a` `b` `c` `d` `e` `f`
+
+OtherPrefixedDocumentIdentifier :: `x-` UnreservedCharacter+ `:`
+UnreservedCharacter+
```
Shane32, line 118: Pursuant to the 7-25-2024 discussion, to allow additional `:` characters within the payload, this definition needs to be updated as it currently only allows unreserved characters in both the prefix and payload.

spec/Appendix A -- Persisted Documents.md
Shane32: > Very strong agree that persisted document IDs should not be hand-crafted. They should be hashes or counters or similar - computer assigned.

I just want to point out that part of the discussion during the 7-25-2024 working group was whether we should add a recommendation to include the document identifier and operation name within the url, specifically to make it easier to be human-readable in logs and during debugging.  Given that many programmers do not utilize operation names, I think it is reasonable to guess that a valid document identifier naming scheme may be a combination of a description and hash/id, similar to machine-generated human-readable SEO product identifiers.

For instance, the algorithm may take the filename of the tsx file containing the query along with 6 base64 characters of the hash (48 bits-worth) for something like `edit_product~3K-A2J-` as the document identifier.

spec/Appendix A -- Persisted Documents.md
martinbonnin: Addressed [here](https://github.com/graphql/graphql-over-http/pull/296/files#diff-9be5577e05ae2112d2b8f95584b162d0dec01453bf6c85df58bf5db4f2c9727aR100)

spec/Appendix A -- Persisted Documents.md
martinbonnin: Addressed [here](https://github.com/graphql/graphql-over-http/pull/296/files#diff-9be5577e05ae2112d2b8f95584b162d0dec01453bf6c85df58bf5db4f2c9727aR121)

Shane32: I like it!

Shane32 (approved):

spec/Appendix A -- Persisted Documents.md
```diff
@@ -95,6 +95,54 @@ implementation specific.
 Note: A 32 character hexadecimal _custom document identifier_ is likely to be an
 MD5 hash of the GraphQL document, as traditionally used by Relay.
 
+### Document identifier syntax
+
+A _document identifier_ must only contain colons (`:`) and characters that are
+defined as
+[`unreserved` in RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3)
+(alphanumeric characters (`A-Z`, `a-z`, `0-9`), dashes (`-`), periods (`.`),
+underscores (`_`), and tildes (`~`)).
```
n1ru4l, line 104: Can we add `/` to the list of allowed characters?

For GraphQL Hive, we already use `"<app_name>/<app_version>/<document_hash>"` as the document id for namespacing documents. (https://the-guild.dev/graphql/hive/docs/features/app-deployments#sending-persisted-document-requests-from-your-app)

From a persisted document store point of view, it does not make sense for us to share the hashes between all the different GraphQL clients.

You might not own all the teams that write to the persisted document store and dictating the hash algorithm to be used for everyone might not make sense.

In addition, we use the extracted app_name and app_version for analytics of the GraphQL API client usage, but that is irrelevant for this request, but might be relevant for the motivations.

martinbonnin: That'd works for me 👍 . 

IIRC, there was a desire that the identifier should be "url-compatible" to use it as part of the url with GET request and improve the debugging/monitoring experience. And `/` might have a specific meaning there. I'm personally fine with letting implementers deal with that.

I'll ultimately defer to @benjie on this one as he has put a lot of thought into this.

Shane32: Already part of the spec:

https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#get

> For HTTP GET requests, the GraphQL-over-HTTP request parameters MUST be provided in the query component of the request URL, encoded in the application/x-www-form-urlencoded format as specified by the [WhatWG URLSearchParams class](https://url.spec.whatwg.org/#interface-urlsearchparams).

benjie: To be clear, the GET requests being recommended as part of #305 are `/graphql/<hash>/<operationname>`, not using query strings. Technically you can have `hash` here have `%2F` in it to escape the `/`, but I think it would be better to stick to URL-safe characters if possible.

Happy to change a `MUST` to a `SHOULD` such that `/` is allowed. E.g. we RECOMMEND that you only use characters in the given set, but you MAY use others if there's a good reason to do so.

Before doing so, though, I'd ask whether using `~` might be a more acceptable character substitute - suitably uncommon, but also URL-safe.

Shane32: > Happy to change a `MUST` to a `SHOULD` such that `/` is allowed

Personally, I'd do this, along with a note about encoding the hash alongside the `/graphql/<hash>/<operationname>` recommendation in #305 .  I probably would not add `/` to the safe list of characters when we are simultaneously recommending `/` be used to separate the hash from the operation name.

So the additional note might read something like this:
> The document id / hash MUST be encoded using percent-encoding as specified by the [WhatWG URLSearchParams class](https://url.spec.whatwg.org/#interface-urlsearchparams).

We should not need to describe how to encode the document id as a query parameter (e.g. `/graphql?documentId=<hash>`) as this is already specified in the spec.

n1ru4l: This made me aware of something else, doesnt `/graphql/<sha>` already conflict with /graphql/stream used by the GraphQL over SSE protocol?

Would in that case it make sense to use a different pattern such as `/graphql/persisted/sha`?

That would still not solve the /operation-name appendix though..

Using tilde is okay, but certainly looks weird.

Shane32: I've been looking at the GraphQL over SSE specification.  I can't find that the url is defined anywhere, besides being used in some samples - ??  Seems to me that when the server receives `Accept: text/event-stream` then it may (not must, in case of validation errors) respond with a SSE stream...

n1ru4l: @Shane32 You are right it is not part of the specification. But the reference implementation `graphql-sse` kind of promotes using `/graphql/stream`: https://the-guild.dev/graphql/sse/recipes

It is out of topic here though, and I am fine with changing `/` to `~`. It might be worth to also ask other vendors that provide a persisted query feature though, since it can imply breaking changes.

benjie: PostGraphile also uses `/graphql/stream` as the endpoint to subscribe to schema change notifications as specified in https://github.com/graphql/graphql-over-http/issues/48

I agree with using a prefix.

spec/Appendix A -- Persisted Documents.md
Shane32: I attempted an implementation of persisted documents today.  The client is using [encodeUriComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) to encode the document Id into the url in a format such as `/graphql?documentId=sha256:abcdef...`.  I immediately found that `:` was encoded as `%3A` due to the potentical conflict with `://`.  In short, `:` is not a url-safe character.  As such, it's somewhat pointless to argue that we should only allow URL-safe characters in the payload when the document id cannot possibly be URL-safe as we have defined `:`, a most-assuredly non-url-safe character, is the separator between a document id's prefix and its payload.

If we wish for the document ids to be url-safe, we should pick another character (such as `.`) as the separator between prefix and payload.  Using `.` would still allow Base64Url (which uses `_` and `-`).  Or we could use another character such as `!` so that `.` is available for custom identifiers that wish to use `.`.

benjie: Good catch @Shane32; `:` sits in that grey area where you don't strictly need to escape it, but it's best practice to do so. We should change that to a URL-safe character.

benjie: I like the `!` suggestion best I think. I think we can change the spec text to that without making existing `:` usage non-compliant - it just becomes a custom ID rather than prefixed ID.

benjie: `*` seems to work with both `URLSearchParams` and `encodeURIComponent`, and I don't think it's normally used in encodings so should be safe... I'll need to look up in the relevant specs to see if it's not meant to be used for some reason.

Shane32: Spec for URLSearchParams:

> The [application/x-www-form-urlencoded percent-encode set](https://url.spec.whatwg.org/#application-x-www-form-urlencoded-percent-encode-set) contains all code points, except the [ASCII alphanumeric](https://infra.spec.whatwg.org/#ascii-alphanumeric), U+002A (*), U+002D (-), U+002E (.), and U+005F (_).

So `~` is probably encoded also.

PR #303, opened 2024-07-26
Add security and compatibility notes

Shane32:
Based on the conversation within the 7-25-2024 working group, I've proposed a non-normative note on security and compatibility.  Feel free to suggest changes to this text.

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -746,7 +746,7 @@ well-formed _GraphQL response_ since the source of the response may not be the
 server but instead some intermediary such as API gateways, proxies, firewalls,
 etc.
 
-# Non-Normative Notes
+# Non-normative notes
 
 ## Security
 
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -746,7 +746,7 @@ well-formed _GraphQL response_ since the source of the response may not be the
 server but instead some intermediary such as API gateways, proxies, firewalls,
 etc.
 
-# Additional Notes
+# Non-Normative Notes
 
 ## Security
 
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -779,10 +779,11 @@ For more detailed security considerations, please refer to
 [RFC 7231](https://tools.ietf.org/html/rfc7231),
 [RFC 6454](https://tools.ietf.org/html/rfc6454), and other relevant RFCs.
 
-## Format Compatibility
+## Request Format Compatibility
 
 Supporting formats not described by this specification, such as XML or Protobuf,
-may conflict with future versions of this specification, as ongoing development
-aims to standardize and ensure the security and interoperability of GraphQL over
-HTTP. For this reason, it is recommended to adhere to the officially recognized
-formats outlined here.
+may have potential conflicts with future versions of this specification as
+ongoing development aims to standardize and ensure the security and
+interoperability of GraphQL over HTTP. While it is recommended to primarily
+adhere to the officially recognized formats outlined here, experimentation with
+other encodings is encouraged to explore their potential benefits.
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -784,5 +784,5 @@ For more detailed security considerations, please refer to
 Supporting formats not described by this specification, such as XML or Protobuf,
 may have potential conflicts with future versions of this specification as
 ongoing development aims to standardize and ensure the security and
-interoperability of GraphQL over HTTP.  For this reason, it is recommended to adhere to the officially recognized
-formats outlined here.
+interoperability of GraphQL over HTTP. For this reason, it is recommended to
+adhere to the officially recognized formats outlined here.
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -745,3 +745,35 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
 well-formed _GraphQL response_ since the source of the response may not be the
 server but instead some intermediary such as API gateways, proxies, firewalls,
 etc.
+
+# Additional Notes
+
+## Security
+
+In this specification, GET requests are not supported for mutations due to
+security concerns. GET requests expose variables to logging mechanisms and
+intermediaries due to the URL encoding of parameters, which can lead to
+sensitive data being inadvertently logged. Furthermore, GET requests are
+considered "basic requests" under CORS (Cross-Origin Resource Sharing),
+meaning they bypass preflight checks that add a layer of security.
+
+On the other hand, using `application/json` for request bodies mandates a CORS
+preflight request, adding a security layer by ensuring the client has explicit
+permission from the server before sending the actual request. This is
+particularly important in mitigating cross-site request forgery (CSRF) attacks.
+
+Additionally, supporting form data requests (`application/x-www-form-urlencoded`
+or `multipart/form-data`) could pose significant security risks. Form data requests
+may be vulnerable to CSRF and other attacks due to the lack of CORS preflight checks.
+As a result, the use of form data for GraphQL queries or mutations is discouraged.
+
+For more detailed security considerations, please refer to [RFC 7231](https://tools.ietf.org/html/rfc7231),
+[RFC 6454](https://tools.ietf.org/html/rfc6454), and other relevant RFCs.
+
+## Format Compatibility
+
+Supporting formats not described by this specification, such as XML or protobuf,
+may conflict with future versions of this specification, as ongoing development
+aims to standardize and ensure the security and interoperability of GraphQL over
+HTTP. For this reason, it is recommended to adhere to the officially recognized
+formats outlined here.
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -762,11 +762,18 @@ preflight request, adding a security layer by ensuring the client has explicit
 permission from the server before sending the actual request. This is
 particularly important in mitigating cross-site request forgery (CSRF) attacks.
 
-Additionally, supporting form data requests (`application/x-www-form-urlencoded`
-or `multipart/form-data`) could pose significant security risks. Form data
-requests may be vulnerable to CSRF and other attacks due to the lack of CORS
-preflight checks. As a result, the use of form data for GraphQL queries or
-mutations is discouraged.
+It's important to note that "simple requests" like those using
+`application/x-www-form-urlencoded` or `multipart/form-data` do not have the
+same CORS behavior, and thus do not undergo the same preflight checks.
+Implementers should be aware of the security implications of using these types
+of requests. While they can be secured with the right headers enforced by the
+server, it is crucial to understand and properly account for the security risks
+involved.
+
+To mitigate these risks, it is recommended that servers require a custom header
+to ensure requests are not "simple." For instance, a `GraphQL-Require-Preflight`
+header can be used to indicate that a preflight check has occurred, providing an
+additional layer of security.
 
 For more detailed security considerations, please refer to
 [RFC 7231](https://tools.ietf.org/html/rfc7231),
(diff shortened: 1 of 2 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,37 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
...
+
+# Additional Notes
+
+## Security
+
+In this specification, GET requests are not supported for mutations due to
+security concerns. GET requests expose variables to logging mechanisms and
+intermediaries due to the URL encoding of parameters, which can lead to
+sensitive data being inadvertently logged. Furthermore, GET requests are
+considered "basic requests" under CORS (Cross-Origin Resource Sharing), meaning
```
jaydenseric, line 757:
```suggestion
considered "simple requests" under CORS (Cross-Origin Resource Sharing), meaning
```

It's referred to as "simple", not "basic":

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,37 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
...
+On the other hand, using `application/json` for request bodies mandates a CORS
+preflight request, adding a security layer by ensuring the client has explicit
+permission from the server before sending the actual request. This is
+particularly important in mitigating cross-site request forgery (CSRF) attacks.
+
+Additionally, supporting form data requests (`application/x-www-form-urlencoded`
+or `multipart/form-data`) could pose significant security risks. Form data
+requests may be vulnerable to CSRF and other attacks due to the lack of CORS
+preflight checks. As a result, the use of form data for GraphQL queries or
+mutations is discouraged.
```
jaydenseric, line 769: Not keen on this wording; `multipart/form-data` can be done securely regarding CORS with the right headers enforced by the server. Lots of people successfully make GraphQL queries and mutations with [GraphQL multipart requests](https://github.com/jaydenseric/graphql-multipart-request-spec) in production systems.

The wording "is discouraged" is hostile to the [GraphQL multipart request spec](https://github.com/jaydenseric/graphql-multipart-request-spec), which one day will be moved here into a GraphQL over HTTP spec.

A better approach would be to mention that "simple requests" don't have normal CORS behavior, and to give a few examples of the kinds of requests that are "simple". `multipart/form-data` shouldn't be singled out as somehow worse than any other kind of simple request. In my spec, I don't prescribe what people should do to achieve security, I just note that they should be aware that they need to understand the implications and account for it:

https://github.com/jaydenseric/graphql-multipart-request-spec?tab=readme-ov-file#security

Most people just enforce in their GraphQL server middleware that if the request is `multipart/form-data`, a certain custom header must be present, indicating that the request isn't "simple".

jaydenseric (changes requested):

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,37 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
 well-formed _GraphQL response_ since the source of the response may not be the
 server but instead some intermediary such as API gateways, proxies, firewalls,
 etc.
+
+# Additional Notes
```
martinbonnin, line 749: [In the main spec](https://spec.graphql.org/draft/#sec-Non-Normative-Portions), all text is normative except otherwise sepcified. Might be worth being explicit?

```suggestion
# Non-Normative Notes
```

spec/GraphQLOverHTTP.md
martinbonnin: +1 Apollo has `Apollo-Require-Preflight` ([doc](https://www.apollographql.com/docs/apollo-server/security/cors/#preventing-cross-site-request-forgery-csrf)). 

I'm not overly familiar with the topic but I would welcome an explicit recommendation in this note to help implementers. `GraphQL-Require-Preflight` maybe?

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,37 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
...
+[RFC 7231](https://tools.ietf.org/html/rfc7231),
+[RFC 6454](https://tools.ietf.org/html/rfc6454), and other relevant RFCs.
+
+## Format Compatibility
+
+Supporting formats not described by this specification, such as XML or Protobuf,
+may conflict with future versions of this specification, as ongoing development
+aims to standardize and ensure the security and interoperability of GraphQL over
+HTTP. For this reason, it is recommended to adhere to the officially recognized
+formats outlined here.
```
martinbonnin, line 781: +1 on the concern but would also like to encourage experimentations with other encodings since JSON is quite ineffective (see also https://github.com/graphql/graphql-spec/issues/432).

We also have language in the current spec that alternative formats are supported ([source](https://github.com/graphql/graphql-over-http/blob/68d029773271278c7cf5ea2545ab303829b0db59/spec/GraphQLOverHTTP.md?plain=1#L167)):

```
Servers and clients MUST support JSON and MAY support other, additional
serialization formats.
```

So the current text somewhat contradicts that.

What about we reserve all `graphql-response+*` to the GraphQL spec and users can use the rest of the namespace for experimentation?

martinbonnin (commented):
Thanks for addressing this! Comments below.

spec/GraphQLOverHTTP.md
Shane32: I the concern relates to request formats moreso than response formats.  Presumably if the client uses a nonstandard request format, they will already have an expectation of the response format.  If there becomes an additional standard to the spec that creates a conflict in the request format, the server may not understand the request or respond in the format the client expects.  So I think the conflict with future request formats is what needs to be addressed here.

martinbonnin: Gotcha 👍 Then maybe make it explicit that this is about requests? 

```md
## Request body media type compatibility

For request bodies, supporting media types not described in this specification, such as XML, ...
```

spec/GraphQLOverHTTP.md
Shane32: @jaydenseric @martinbonnin I took another pass at this paragraph.  Please review the revised PR - thanks!

spec/GraphQLOverHTTP.md
Shane32: I've taken another pass at this paragraph.  Please review the revised PR - thanks!

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,45 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
...
+same CORS behavior, and thus do not undergo the same preflight checks.
+Implementers should be aware of the security implications of using these types
+of requests. While they can be secured with the right headers enforced by the
+server, it is crucial to understand and properly account for the security risks
+involved.
+
+To mitigate these risks, it is recommended that servers require a custom header
+to ensure requests are not "simple." For instance, a `GraphQL-Require-Preflight`
+header can be used to indicate that a preflight check has occurred, providing an
+additional layer of security.
```
martinbonnin, line 776: 👍

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,45 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
...
+[RFC 6454](https://tools.ietf.org/html/rfc6454), and other relevant RFCs.
+
+## Request Format Compatibility
+
+Supporting formats not described by this specification, such as XML or Protobuf,
+may have potential conflicts with future versions of this specification as
+ongoing development aims to standardize and ensure the security and
+interoperability of GraphQL over HTTP. While it is recommended to primarily
+adhere to the officially recognized formats outlined here, experimentation with
+other encodings is encouraged to explore their potential benefits.
```
martinbonnin, line 789: My other comment was mainly about response formats. For request formats TBH I'm not sure if I want to encourage that since I'm not sure what the benefits are (probably a little benefit but a lot less than for responses) so I'm fine with your previous wording:

```suggestion
interoperability of GraphQL over HTTP.  For this reason, it is recommended to adhere to the officially recognized
formats outlined here.
```

We could also explore request content-types like `graphql-request+json` if that hasn't already been done. But that's a problem for another day!

Either way, both options work for me, I'll leave it out to you.

martinbonnin (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,45 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
...
+To mitigate these risks, it is recommended that servers require a custom header
+to ensure requests are not "simple." For instance, a `GraphQL-Require-Preflight`
+header can be used to indicate that a preflight check has occurred, providing an
+additional layer of security.
+
+For more detailed security considerations, please refer to
+[RFC 7231](https://tools.ietf.org/html/rfc7231),
+[RFC 6454](https://tools.ietf.org/html/rfc6454), and other relevant RFCs.
+
+## Request Format Compatibility
```
Shane32, line 782:
```suggestion
## Request format compatibility
```

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,45 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
 well-formed _GraphQL response_ since the source of the response may not be the
 server but instead some intermediary such as API gateways, proxies, firewalls,
 etc.
+
+# Non-Normative Notes
```
Shane32, line 749:
```suggestion
# Non-normative notes
```

spec/GraphQLOverHTTP.md
```diff
@@ -746,7 +746,7 @@ well-formed _GraphQL response_ since the source of the response may not be the
 server but instead some intermediary such as API gateways, proxies, firewalls,
 etc.
 
-# Non-Normative Notes
+# Non-normative notes
```
martinbonnin: Thanks for that 🙏 I was looking at the rendered view. Didn't realize the translation was made automagically 👍

spec/GraphQLOverHTTP.md
Shane32: :+1: Suggestion above is merged.

spec/GraphQLOverHTTP.md
Shane32: @jaydenseric Any comments on the revised PR?

JoviDeCroock (approved):

martinbonnin (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,44 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
...
+## Security
+
+In this specification, GET requests are not supported for mutations due to
+security concerns. GET requests expose variables to logging mechanisms and
+intermediaries due to the URL encoding of parameters, which can lead to
+sensitive data being inadvertently logged. Furthermore, GET requests are
+considered "simple requests" under CORS (Cross-Origin Resource Sharing), meaning
+they bypass preflight checks that add a layer of security.
+
+On the other hand, using `application/json` for request bodies mandates a CORS
```
alf, line 760: This is a very nice addition.

The term “simple request” is used but not defined and I can’t find the definition in any of the references either. How about defining it using the following quote from the fetch spec? I believe that’s the authoritative source:

> For [requests](https://fetch.spec.whatwg.org/#concept-request) that are more involved than what is possible with HTML’s [form](https://html.spec.whatwg.org/multipage/forms.html#the-form-element) element, a [CORS-preflight request](https://fetch.spec.whatwg.org/#cors-preflight-request) is performed, to ensure [request](https://fetch.spec.whatwg.org/#concept-request)’s [current URL](https://fetch.spec.whatwg.org/#concept-request-current-url) supports the [CORS protocol](https://fetch.spec.whatwg.org/#cors-protocol).

https://fetch.spec.whatwg.org/#http-cors-protocol

spec/GraphQLOverHTTP.md
```diff
@@ -745,3 +745,44 @@ payload is `application/json` then the client MUST NOT rely on the body to be a
...
+
+It's important to note that "simple requests" like those using
+`application/x-www-form-urlencoded` or `multipart/form-data` do not have the
+same CORS behavior, and thus do not undergo the same preflight checks.
+Implementers should be aware of the security implications of using these types
+of requests. While they can be secured with the right headers enforced by the
+server, it is crucial to understand and properly account for the security risks
+involved.
+
+To mitigate these risks, it is recommended that servers require a custom header
```
alf, line 773: How about:

To mitigate these risks, it is recommended that servers require a custom header. Since this is not possible using forms, a preflight check must occur. While the name and value of the header is not important, we suggest `GraphQL-Require-Preflight`.

PR #304, opened 2024-07-27
Attempt at clarifying server vs intermediary service

martinbonnin:
Comments inline

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -511,19 +511,16 @@ execution regardless of validation errors.
 
 ## Status Codes
 
-In case of errors that completely prevent the generation of a well-formed
-_GraphQL response_, the server SHOULD respond with the appropriate status code
-depending on the concrete error condition, and MUST NOT respond with a `2xx`
-status code when using the `application/graphql-response+json` media type.
+The status codes depends on the media type with which the GraphQL response will
+be served.
 
-Note: Typically the appropriate status code will be `400` (Bad Request).
-
-Note: This rule is "should" to maintain compatibility with legacy servers which
-can return 200 status codes even when this type of error occurs, but only when
-not using the `application/graphql-response+json` media type.
-
-Otherwise, the status codes depends on the media type with which the GraphQL
-response will be served:
+Note: when using the `application/graphql-response+json` media type, the status
+code is primarily aimed for intermediary servers that do not understand GraphQL.
+In these cases, clients can rely on the response being a well-formed _GraphQL
+response_ regardless of the status code. When using the `application/json` media
+type on the other hand, clients may use a `2xx` status code as an indication
+that the body contains a well-formed _GraphQL response_. See
+[processing a response](#sec-Processing-a-response) for more details.
 
 ### application/json
 
@@ -738,10 +735,15 @@ Note: The GraphQL specification
 and refers to the situation wherein a _GraphQL field error_ occurs as a partial
 response; it still indicates successful execution.
 
-## Processing the response
+## Processing a response
 
-If the response uses a non-`200` status code and the media type of the response
-payload is `application/json` then the client MUST NOT rely on the body to be a
-well-formed _GraphQL response_ since the source of the response may not be the
-server but instead some intermediary such as API gateways, proxies, firewalls,
-etc.
+In a typical environment, the source of a response may be some intermediary
+server such as an API gateway, a proxy, a firewall, etc.
+
+Those intermediary servers may not understand GraphQL: in the case of an error,
+they may return their own non-GraphQL `application/json` response with a
+non-`2xx` status code.
+
+For this reason, if the response uses a non-`2xx` status code and the media type
+of the response body is `application/json` then the client MUST NOT rely on the
+body to be a well-formed _GraphQL response_.


spec/GraphQLOverHTTP.md
```diff
@@ -511,19 +511,16 @@ execution regardless of validation errors.
 
 ## Status Codes
 
-In case of errors that completely prevent the generation of a well-formed
-_GraphQL response_, the server SHOULD respond with the appropriate status code
-depending on the concrete error condition, and MUST NOT respond with a `2xx`
-status code when using the `application/graphql-response+json` media type.
```
martinbonnin, line 517: IMO, there's a tension between the [Body first paragraph](https://graphql.github.io/graphql-over-http/draft/#sec-Body):

```
The body of the server’s response MUST follow the requirements for a GraphQL response
encoded directly in the chosen media type.
```

and that paragraph, which is why I removed it. 

My current understanding is that this paragraph was meant for intermediary servers that may not understand GraphQL. So there are 2 kind of servers:

* GraphQL server
* Intermediary server

```
Client <--> [Intermediary Server] <--> GraphQL Server
```

I'd say this spec should be about GraphQL servers and not Intermediary servers? Or if we want to specify the behaviour of intermediary server, we could do it in a separate section?

Or were there use cases when a GraphQL server is compliant and does not return a well formed GraphQL response?

spec/GraphQLOverHTTP.md
```diff
@@ -511,19 +511,16 @@ execution regardless of validation errors.
...
-
-Otherwise, the status codes depends on the media type with which the GraphQL
-response will be served:
+Note: when using the `application/graphql-response+json` media type, the status
+code is primarily aimed for intermediary servers that do not understand GraphQL.
+In these cases, clients can rely on the response being a well-formed _GraphQL
+response_ regardless of the status code. When using the `application/json` media
+type on the other hand, clients may use a `2xx` status code as an indication
+that the body contains a well-formed _GraphQL response_. See
+[processing a response](#sec-Processing-a-response) for more details.
```
martinbonnin, line 523: This paragraph mostly duplicates `6.5 processing a response` but is added for discoverability and also expliciting the role of the status code.

Shane32, line 523: This seems confusing to me.

> Note: when using the `application/graphql-response+json` media type

To me this implies that the response is the above type

> the status code is primarily aimed for intermediary servers that do not understand GraphQL

In which case the status code was provided by the GraphQL server.

But I am guessing the intent here is more like "when expecting the `application/graphql-response+json` media type", then this and that.

On a side note, I would say the status code is always "for" the client, not "for" intermediary servers.

spec/GraphQLOverHTTP.md
```diff
@@ -511,19 +511,16 @@ execution regardless of validation errors.
 
 ## Status Codes
 
-In case of errors that completely prevent the generation of a well-formed
-_GraphQL response_, the server SHOULD respond with the appropriate status code
-depending on the concrete error condition, and MUST NOT respond with a `2xx`
-status code when using the `application/graphql-response+json` media type.
+The status codes depends on the media type with which the GraphQL response will
+be served.
```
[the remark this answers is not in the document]
martinbonnin: Completely agree, I just didn't know what to put as a lead-in paragraph and went for the "easy-but-out-of-place" option. I'll reword.

spec/GraphQLOverHTTP.md
martinbonnin: > To me this implies that the response is the above type

👍  I'll reword

> I would say the status code is always "for" the client, not "for" intermediary servers.

This has been my expectation for the last 7 years 😅. I agree that status code in general are meant for clients but the more I look at the `graphql-response+json` new way to do things, the more I feel I have been wrong all this time... (See also [Discord conversation](https://discord.com/channels/625400653321076807/863141924126588958/1266313412585979955)). 

If I'm a client and I'm receiving a `graphql-response+json` response, I'll probably not even look at the status code and just forward everything to the user.

spec/GraphQLOverHTTP.md
```diff
@@ -738,10 +735,15 @@ Note: The GraphQL specification
 and refers to the situation wherein a _GraphQL field error_ occurs as a partial
 response; it still indicates successful execution.
 
-## Processing the response
+## Processing a response
+
+In a typical environment, the source of a response may be some intermediary
+server such as an API gateway, a proxy, a firewall, etc.
```
Shane32, line 741: This appears to imply that typically the response comes from an intermediary.  I would disagree.  I think we should remove "typical" and reword the sentence so it flows better. Rough draft:

```suggestion
The source of a response may be an intermediary server, such as an API gateway, proxy, or
firewall, in certain environments.
```

spec/GraphQLOverHTTP.md
```diff
@@ -738,10 +735,15 @@ Note: The GraphQL specification
...
+
+Those intermediary servers may not understand GraphQL: in the case of an error,
+they may return their own non-GraphQL `application/json` response with a
+non-`2xx` status code.
```
Shane32, line 745: I don't see the relevancy of whether the intermediary servers understand GraphQL.  Perhaps we could say that they 'may not return a GraphQL response', but then, what servers would?  Ultimately anything that returns a GraphQL-formatted response, such as a Federation router, is probably the API endpoint itself, and so this comment does not really apply as the response is being handled there, and not passed opaquely back to the client.

martinbonnin: Agree for a Federation router this sentence makes not sense but what about a CDN? 

Let's say you put your GraphQL server behind Cloudflare using persisted documents so everything that Cloudflare sees is REST (don't ask me why anyone would do that, I'm just a poor client developer but let's assume it's a thing). 

If Cloudflare detects that a botnet is using your endpoint, they're going to reply with:

```
429 Too many
Content-type: application/json

{
  "message": "oh no we detected a botnet",
  "..."
}
```

In the 2xx case, we're sure the source of the response is the GraphQL server but not in the other cases? At least this is my understanding of it.

spec/GraphQLOverHTTP.md
Shane32: At first glance, it appears the paragraph was intended to address invalid requests.  For example, if the request was not able to be parsed as JSON or had an unrecognized content type.  This paragraph would enforce that a non-200 status code was used.  Whereas if we read just above the 6.1 paragraph you referenced, paragraph 6 starts with:

> When a server receives a well-formed [GraphQL-over-HTTP request](https://graphql.github.io/graphql-over-http/draft/#graphql-over-http-request) ...

martinbonnin: My gripe is specifically about that part: `In case of errors that completely prevent the generation of a well-formed
_GraphQL response_`. That contradicts 6.1. I understand to be 6.1 more general about the body. Then the status codes are refined below.

 If the request was not able to be parsed as JSON or had an unrecognized content type, I think there are other places in the spec that address this already: 

*  request was not able to be parsed as JSON
  * [application/json case](https://graphql.github.io/graphql-over-http/draft/#sel-FANNLRCAACEBi8K) 
  * [graphql-response+json case](https://graphql.github.io/graphql-over-http/draft/#sel-FANNNRCAACENz5F)
* [unrecognized content type](https://graphql.github.io/graphql-over-http/draft/#sel-EANHJCAACElBzyf)

spec/GraphQLOverHTTP.md
martinbonnin: I have reworded this to remove the "primarily aimed". We don't know how people are going to use it so I went for something more neutral.

spec/GraphQLOverHTTP.md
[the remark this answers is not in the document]
martinbonnin: I've reworked this in https://github.com/graphql/graphql-over-http/pull/304/commits/86819115b6eef6b98d1dfdf4f26fa04b6c379c14. Didn't get much inspiration about how to generally describe the status code so I ended up splitting the note in 2, one for each section.

spec/GraphQLOverHTTP.md
Shane32: Correct.  I was just saying that whether the CDN "understands" GraphQL doesn't really make sense.  The CDN is a caching service.  A proxy or firewall is a relay service.  They don't parse GraphQL requests; if they did, such as in Apollo Federation, they are essentially the endpoint themselves and the GraphQL server is just responding to requests from the firewall.  The goal here is to just ensure that a 400 application/json response code from the GraphQL server can be distinguished from a 400 application/json response code from an intermediary server.

martinbonnin: >  I was just saying that whether the CDN "understands" GraphQL doesn't really make sense. 

The sentence is informative only. Just highlights that there might be non-GraphQL links in the chain, which you might forget if you're reading the spec from a client point of view. Agree it's technically not needed but I don't think it hurts?

> The goal here is to just ensure that a 400 application/json response code from the GraphQL server can be distinguished from a 400 application/json response code from an intermediary server.

Ah! This might be what I'm missing there. Is that possible based on just `content-type` + `status code`? In the general case, a CDN error could potentially clash with any GraphQL error response?

spec/GraphQLOverHTTP.md
```diff
@@ -671,6 +665,12 @@ pass validation, then the server SHOULD reply with `400` status code.
 If the client is not permitted to issue the GraphQL request then the server
 SHOULD reply with `403`, `401` or similar appropriate status code.
 
+Note: When the response media type is `application/graphql-response+json`,
+clients can rely on the response being a well-formed _GraphQL response_
+regardless of the status code. The intermediary servers that do not understand
+GraphQL may use the status code to get some information about the shape of the
+_GraphQL response_.
```
Shane32, line 672: Good non-normative note. :+1:

spec/GraphQLOverHTTP.md
```diff
@@ -568,6 +557,11 @@ of `2xx` or `5xx` status codes when responding to invalid requests using the
 Note: URLs that enable GraphQL requests may enable other types of requests - see
 the [URL](#url) section.
 
+Note: When the response media type is `application/json`, clients may use a
+`2xx` status code as an indication that the body contains a well-formed _GraphQL
+response_. See [processing a response](#sec-Processing-a-response) for more
+details.
```
Shane32, line 563: This seems to essentially mirror the first paragraph of this section.  Perhaps a more useful non-normative note would be a negative example, such as:

```suggestion
Note: When the response media type is `application/json`, clients cannot count on
the response being a well-formed _GraphQL response_ unless the `2xx` status code
is also returned. See [processing a response](#sec-Processing-a-response) for more
details.
```

spec/GraphQLOverHTTP.md
Shane32: Yeah, so like the `application/json` + 400 could be CloudFlare's response, or it could be a GraphQL response.  So we recommend that 200 is returned for `application/json` responses even in the case of an error.  Or, use `application/graphql-response+json` so that the response is known to be able to be parsed by the client, even in the even of a 400 error.

spec/GraphQLOverHTTP.md
Shane32: Or:

```suggestion
Note: When the response media type is `application/json` and the status code is 4xx,
clients cannot count on the response being a well-formed _GraphQL response_.
See [processing a response](#sec-Processing-a-response) for more details.
```

spec/GraphQLOverHTTP.md
martinbonnin: Gotcha 👍 Makes sense. I think this is covered [here](https://graphql.github.io/graphql-over-http/draft/#sel-FANNLHCCBCOwDl2O)

spec/GraphQLOverHTTP.md
martinbonnin: I went with your last solution with a [few tweaks](https://github.com/graphql/graphql-over-http/pull/304/commits/f73fd2c5a59714ff5a1b4247496b019e13b2f104)

FWIW, what I'm trying to do in this note is make the parsing of the section easier from a client point of view because the real information is in section 6.1 (body) but it's easy to come to section 6.4 to know how to interpret the status code.

martinbonnin: Rereading the re-rendered document, I think there is some repetition indeed so I backtracked and just [added the "processing a response" link to the first note](https://github.com/graphql/graphql-over-http/pull/304/commits/110d6adab7b832d3d2ba88fe72ac8279a636907e)

spec/GraphQLOverHTTP.md
Shane32: I'm not sure this paragraph contradicts anything, but I'm okay removing it as redundant / described elsewhere.

spec/GraphQLOverHTTP.md
```diff
@@ -738,10 +734,15 @@ Note: The GraphQL specification
...
 
-## Processing the response
+## Processing a response
+
+The source of a response may be an intermediary server, such as an API gateway,
+proxy, or firewall, in certain environments.
+
+Those intermediary servers may not understand GraphQL: in the case of an error,
+they may return their own non-GraphQL `application/json` response with a
+non-`200` status code.
```
Shane32, line 744: I'd still recommend removing the 'may not understand GraphQL' part.  I'm good with the rest of this PR.

```suggestion
In certain environments, the source of a response may be an intermediary server,
such as an API gateway, proxy, or firewall. In the case of an error, they may return
their own non-GraphQL `application/json` response with a non-`200` status code.
```

spec/GraphQLOverHTTP.md
```diff
@@ -738,10 +734,15 @@ Note: The GraphQL specification
...
 
-If the response uses a non-`200` status code and the media type of the response
-payload is `application/json` then the client MUST NOT rely on the body to be a
-well-formed _GraphQL response_ since the source of the response may not be the
-server but instead some intermediary such as API gateways, proxies, firewalls,
-etc.
+For this reason, if the response uses a non-`200` status code and the media type
+of the response body is `application/json` then the client MUST NOT rely on the
+body to be a well-formed _GraphQL response_.
```
Shane32, line 745: I just want to point out that sentence defines a rule that the client must follow.  The specification is a server specification, correct?  Do we have any other client requirements within this spec?  Perhaps section this should be a non-normative note as to client behavior.

martinbonnin: Agreed, the client sentences feel a bit out of place. I'll update that.

spec/GraphQLOverHTTP.md
martinbonnin: The contradiction IMO is in the spec saying: `the body must always be a well formed GraphQL response`. And then going on to say "In case of errors that prevent the generation of a well-formed _GraphQL response_", which I understand as `if it isn't a well formed GraphQL response, then ...`

How can it **not** be a well formed GraphQL response if it must always be a well formed GraphQL response? 

The only reason I see for the body to not be a well formed GraphQL response is if content negociation fails. But as soon as `application/json` or `graphql-response+json` is selected, my reading is that the body must always be a well formed GraphQL response? 

If content-negociation errors are what's at stake here (i.e. "the errors that prevent the generation of a well formed GraphQL response"), then this case is already handled by 6.1 so I'd remove the sentence. If there are other error cases, I'd love to explicit them cause it's not obvious what they could be.

spec/GraphQLOverHTTP.md
martinbonnin: Updated [here](https://github.com/graphql/graphql-over-http/pull/304/commits/ddd4ba9ab9721b53b02faf49a08a8ac5f554e829). I also took this opportunity to remove the spec-like language (`MUST NOT rely on ...`) and replace it with pseudo-logic that I would implement in a client application. Let me know what you think.

Shane32: Looks good to me.  Normally I might write 'can only rely on' or something (with the word 'only') but that would preclude optional additional behaviors on the part of the server.

Shane32 (approved):

spec/GraphQLOverHTTP.md
martinbonnin: I contemplated that 'only' for a while too and decided to leave it out ultimately for the same reason. What about adding a trailing sentence:

```
The other cases require the client to have knowledge about the specific service behavior 
and are outside the scope of this specification.
```
?

Shane32: Eh.  Whatever you think.  I'd probably skip it.

martinbonnin: Sounds good, let's keep it simple 👍 . Thanks for all the feedbacks 🙏

spec/GraphQLOverHTTP.md
benjie: > The body of the server’s response MUST follow the requirements for a GraphQL response encoded directly in the chosen media type.

I think this was poorly worded; perhaps:

> When encoding a GraphQL response, the body of the server’s response MUST follow the requirements in the chosen media type.

Essentially it's saying _when you encode a GraphQL response_ that it _must adhere to the spec_. The other paragraph (the one you've deleted) is saying _if_ you cannot form a valid GraphQL response _then_ you must not use a GraphQL response type.

---

> and MUST NOT respond with a `2xx` status code when using the `application/graphql-response+json` media type.

This is also incorrect; more correct would be:

> and MUST NOT respond with the `application/graphql-response+json` media type.

The `application/graphql-response+json` is only for use with well-formed GraphQL responses, so if you can't form a well-formed GraphQL response you can't use it. I think I was probably thinking something like: `try { return doTheGraphQLStuff() } catch (e) { return { errors: [{message: e.message}] }}` where you could then use the graphql-response+json for these errors; but this _is_ a well-formed GraphQL response so it contradicts the beginning of the sentence.

---

This doesn't really refer to "intermediary" servers as much as it does the hosting environment. For example if you use `nginx` and `lua` to do your GraphQL, but the `lua` process crashes, then this requires that your `nginx` isn't configured to still use the `application/graphql-response+json` response type.

martinbonnin: > Essentially it's saying when you encode a GraphQL response that it must adhere to the spec. The other paragraph (the one you've deleted) is saying if you cannot form a valid GraphQL response then you must not use a GraphQL response type.

This is what my brain has a really hard time processing  😅 . Thinking in  terms of logic:

proposition P: `content-type == application/graphql-response+json` 
proposition Q: `isWellFormGraphQLResponse(body) == true`

> when you encode a GraphQL response that it must adhere to the spec

I think this is the same as `if P then Q`?

> if you cannot form a valid GraphQL response then you must not use a GraphQL response type.

I understand map that as `if !Q then !P`?

So the second sentence is the contrapositive of the first one and therefore strictly equivalent from a logical point of view? 

If it brings the attention to a specific point I'm fine with it as a non-normative note but from in the spec text itself, I'd expect no redundancy. Or is there some nuance there?

benjie: > > when you encode a GraphQL response that it must adhere to the spec

> I think this is the same as `if P then Q`?

I don't think so. `P` as you have stated it is that you use a specific media type, but what's being stated is that you are encoding a GraphQL response; it could be in any supported media type.

martinbonnin: Alright 👍  But I think my initial point that it's not clear whether the "non-valid GraphQL response" is part of the spec or not still holds. 

I put the paragraph back in to keep that PR small and about non-normative client notes. Will open a follow up one for "non-valid GraphQL responses"

martinbonnin:
@Shane32 @benjie I have updated that PR to only be about "clients" and non-normative notes. It's additive only now. 

I'll look at the "non-valid GraphQL responses" in a separate PR. Let me know what you think!

Shane32 (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -738,10 +745,17 @@ Note: The GraphQL specification
...
+code.
+
+For this reason, a client application can rely on the response being a
+well-formed _GraphQL response_ if any of the following cases is true:
 
-If the response uses a non-`200` status code and the media type of the response
-payload is `application/json` then the client MUST NOT rely on the body to be a
-well-formed _GraphQL response_ since the source of the response may not be the
-server but instead some intermediary such as API gateways, proxies, firewalls,
-etc.
```
benjie, line 747: The removal of this *normative* paragraph is, in my opinion, wrong. The client must not rely on it being a GraphQL response; if they want to treat it as a GraphQL response they must perform their own validations in order to do so (e.g. assert that it conforms to the correct shape/types), and must know that what they're doing is outside of the spec.

benjie (changes requested):

spec/GraphQLOverHTTP.md
Shane32: @benjie , the way I see it, this is a server spec, not a client spec.  A server has no control over whether or not a client processes the response correctly, incorrectly, or at all.  I think that's why this should be a non-normative paragraph.  And the modified language is, I feel, much easier to understand.

Now we could write a client spec, defining what features a client must and must not have.  For example, we might require the client send the request with a specific content type, or perhaps to support automatic fallback to a known content type if a server responds with an unsupported media type error.  We might define the exact mechanism by which the response shape is verified for known response content types, etc.  While it seems much of this is implicit based on the server requirements, the existing specification is not currently written as a client spec and does not contain any requirement written from the perspective of the client (apart from the little written in this paragraph).  As such, I feel it is improper for it to have a normative note about processing a response in this spec.

If this paragraph were to be kept as a normative paragraph, I feel it should at minimum be in a section labeled for client requirements.  But even then, the existing paragraph says "MUST NOT rely on the body", it doesn't explain what TO do in that situation.  Should it look for a certain shape?  Should it ignore all data in the response?  In other words, it's says "client MUST NOT count on it being well-formed, but it MIGHT be", which is not particularly helpful for a normative note.

What do you think?

martinbonnin: Apologies for the delay here.  +1 to @Shane32 comment.

As a client developer relatively new to this spec but pretty familiar with GraphQL, it took me way longer than I'd like to admit to understand the nuances there. 

Having the language from the point of view of the client is easier to process in my opinion.

benjie: > this is a server spec, not a client spec

That does not seem accurate. This spec specifies:

- how a client should form a request including what details the client must put in the request for it to be likely to be executed
- how a server should check these details, and handle when the client has not conformed
- how a server should form a response including what details to include
- how and when a client should process a response, how to know it originated from the GraphQL service or not (and thus whether it can "trust" that the response is GraphQL or not)

It's a protocol spec that defines the protocol for communicating between a spec-compliant client and server. I believe @enisdenjo implemented both a client and a server based on this spec in `graphql-http`.

> A server has no control over whether or not a client processes the response correctly, incorrectly, or at all.

Indeed; but the client does; hence why we state:

> the client MUST NOT rely on the body to be a
> well-formed _GraphQL response_ since the source of the response may not be the
> server but instead some intermediary such as API gateways, proxies, firewalls,
> etc.

It is the client's responsibility to handle this concern. The server may not even be the one _supplying_ this response (it may have come from a gateway, proxy, etc), so of course the server cannot influence this - it is entirely the client's responsibility.

> the existing specification is not currently written as a client spec and does not contain any requirement written from the perspective of the client (apart from the little written in this paragraph)

It has an entire chapter dedicated to how to form requests to the server. It also has loads of explicit conformance requirement statements for the client:

> **The role of a [client](https://graphql.github.io/graphql-over-http/draft/#client)** is to issue HTTP requests to a [server](https://graphql.github.io/graphql-over-http/draft/#server) in order to interact with a [GraphQL service](https://spec.graphql.org/draft/#sec-Overview).

> Servers and **clients MUST** support JSON and MAY support other, additional serialization formats.

> **A client SHOULD** indicate the media types that it supports in responses using the Accept HTTP header as specified in [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).

> If **the client** supplies an Accept header, **the client SHOULD** include the media type application/graphql-response+json in the Accept header.

> Before 2025-01-01T00:00:00Z, if **the client** supplies an Accept header, the header **SHOULD** include the application/json media type. After this watershed, this is no longer necessary.

> It is **RECOMMENDED that a client** set the Accept header to application/graphql-response+json; charset=utf-8, application/json; charset=utf-8.

> A **client MUST** indicate the media type of a request body using the Content-Type header as specified in [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).

There are many more.

benjie: > If this paragraph were to be kept as a normative paragraph, I feel it should at minimum be in a section labeled for client requirements.

We don't currently label client and server requirements separately since both are intermingled; such as:

> **A client MUST** indicate the media type of a request body using the Content-Type header as specified in [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
> 
> **A server MUST** support POST requests encoded with the application/json media type (as indicated by the Content-Type header) encoded with UTF-8.
> 
> For POST requests using an officially recognized GraphQL Content-Type without indicating an encoding, the **server MUST** assume the encoding is utf-8.
> 
> If the client does not supply a Content-Type header with a POST request, the **server SHOULD** reject the request using the appropriate 4xx status code.

To do this in one place but not the rest of the spec would be a little strange. And doing it across the spec would make the spec a lot less readable, in my opinion.

> But even then, the existing paragraph says "MUST NOT rely on the body", it doesn't explain what TO do in that situation. Should it look for a certain shape? Should it ignore all data in the response?

Indeed; the client must not rely on the body. What it does with the body is up to it, but it cannot be trusted. Throwing an error is the most reliable way of handling it; but we're not going to require that (though we could RECOMMEND it, I suppose). It could also choose to run its own validations, but at that point it's up to the client to figure out what those should be - we've already told it that the response cannot be trusted, so if it feels it should push ahead anyway, that's its responsibility to figure out.

martinbonnin: > The server may not even be the one supplying this response (it may have come from a gateway, proxy, etc)

This is the main thing IMO.  This should not be a "client" or "server" spec, this is a "wire format" spec.

For convenience, I like `client` and `server` because it's so much used everywhere but really it's only about  "initiating peer" and "responding peer". 

Feels to me that `proxies`, `gateways`, etc.. are outside the spec. Of course, it's useful for context and understanding why certain decisions were made (`graphql-response+json`) but really I see those considerations as "non-normative".

If we're saying there are possibly 3 peers in the spec now (`client`, `server`, `proxy`) then we should define them upfront. (But I'd really prefer that there are only 2 as all the other wire format specification I know about have only 2 peers).


>>  the client MUST NOT rely on the body to be a
>> well-formed GraphQL response since the source of the response may not be the
>> server but instead some intermediary such as API gateways, proxies, firewalls,
>> etc.
> 
> It is the client's responsibility to handle this concern. The server may not even be the one supplying this response (it may have come from a gateway, proxy, etc), so of course the server cannot influence this - it is entirely the client's responsibility.

From the client POV, what not to do is not important. It's impossible to enumerate all the things that a client MUST NOT do:
* A client must not asssume `application/text` responses to contain a valid GraphQL response
* A client must not assume the `Cache-Control` header to be set
* etc... 

The list goes on. What's important is what to do. Everything else is outside the spec.

> What it does with the body is up to it, but it cannot be trusted. 

Everything that's not said in the spec cannot be trusted. No need to add additional language about it. Absence of language means "cannot be trusted".

benjie: > This is the main thing IMO. This should not be a "client" or "server" spec, this is a "wire format" spec.

It's not this though; it's the conformance requirements need to write a compliant client and server, _and_ the wire format that they use between them. Clients have to live in the real world where any arbitrary intermediary can intercept the result and return some nonsense; and it's the graphql-over-http spec's duty to inform them when they can and cannot reasonably trust the response.

> If we're saying there are possibly 3 peers in the spec now (client, server, proxy) then we should define them upfront. (But I'd really prefer that there are only 2 as all the other wire format specification I know about have only 2 peers).

I see what you're saying; but it's not `client`, `server` and `proxy`. It's `client`, `server` and `HTTP connection made over unreliable network layer that may or may not exist between them from time to time with or without various intermediaries` (which I'll call `HTTP` for short). I.e. the third party to this is the HTTP connection happening over the shared network (which consists of routers, switches, proxies, firewalls, hotspots, middleware, authz/authn, enterprise network concerns, IDS, governmental intermediaries, and everything else). We cannot enforce any rules on `HTTP`; all we can do is tell the client and server how to talk to `HTTP` and what data received from `HTTP` should be rejected versus accepted - i.e. how to do **GraphQL** when communicating **over** an **HTTP** network connection. I think we call out this third party, `HTTP`, not just in the name of the spec but also throughout the spec.

> Everything else is outside the spec.

I very much feel that this particular rule should not be outside of the spec. It's incredibly important, and the abundance of "200 is not okay" talks at conferences over the past 9 years I think adds a lot of evidence that this rule in particular is necessary.

martinbonnin: > it's the conformance requirements need to write a compliant client and server, and the wire format that they use between them.

Fair enough 👍

> Clients have to live in the real world where any arbitrary intermediary can intercept the result and return some nonsense

Same is true for every networking protocol. HTTP doesn't warn about a TCP router swapping bytes around. TCP doesn't warn about an ethernet card rewriting bytes, etc... It's a given.

> We cannot enforce any rules on HTTP

We kind of do though? What I understand from the spec is that we're "allocating the `graphql-response+json` media type": if `HTTP` returns a `graphql-response+json` response then it MUST be a valid GraphQL response. I think we expect that from the `HTTP` layer? It's a given.

> I very much feel that this particular rule should not be outside of the spec. It's incredibly important, the abundance of "200 is not okay" talks at conferences over the past 9 years I think adds a lot of evidence that this rule in particular is necessary.

Agree that this must be called out in the spec. Just I wouldn't make it normative as it's already contained in the other parts. Maybe a separate paragraph explaining the rationale for `graphql-response+json`? Basically [this note](https://github.com/graphql/graphql-over-http/blame/main/spec/GraphQLOverHTTP.md#L465-L472) but elevating it to a more preminent position in the document and explaining about client processing there?

benjie: > HTTP doesn't warn about a TCP router swapping bytes around.

As a counter-example, HTTP details a `Content-Length` header and states:

>    When a Content-Length is given in a message where a message-body is
>    allowed, its field value MUST exactly match the number of OCTETs in
>    the message-body. **HTTP/1.1 user agents MUST notify the user when an**
>    **invalid length is received and detected.**

This effectively protects the client against the connection being terminated (as would be expected by `Connection: close`) before all the data is received - one of the expected failure modes for HTTP.

Protocol generally have to somewhat trust the underlying protocol (which is why we have not detailed how HTTP works, and why HTTP doesn't detail how TCP works); however they also need to take into account how, even in a trusted network, they may fail - especially when the failure modes can cause ambiguity.

Our failure mode here is that there's an intermediary that bails out the request early (permissions, failed cache, netsplit, timeout, etc etc etc) and returns a response using a common media type `application/json`. We trust these intermediaries to never do this with a 2xx status code; just as we trust them to not use `application/graphql-response+json` if they're not speaking GraphQL. But we must convey this important information to the client implementers: if they get a non-2xx and it uses this common media type then they must not trust it to be GraphQL even if they trust the underlying network - this is an expected failure mode.

I see what you mean about it effectively being covered by other parts of the spec, but I'm not convinced that it doesn't bear repeating - what problem are we trying to solve by removing this or demoting it to a non-normative note?

martinbonnin: > Our failure mode here is that there's an intermediary that bails out the request early (permissions, failed cache, netsplit, timeout, etc etc etc) and returns a response using a common media type application/json. We trust these intermediaries to never do this with a 2xx status code; just as we trust them to not use application/graphql-response+json if they're not speaking GraphQL. 

I love this paragraph. I would love to see that kind of context added to the spec (permissions, failed cache, netsplit, timeout, etc...trust that if this happens it uses non-2xx, etc..)

> they must not trust it to be GraphQL even if they trust the underlying network - this is an expected failure mode.

When I read that as a client developer, "I MUST NOT trust something", I'm not sure what this concretely mean. If I get a `2xx` or a `graphql-response+json` then I know what to do (parse the GraphQL response), if not then 🤷 I must not trust it but how does that concretely materialize? I'm not sure.

> what problem are we trying to solve by removing this or demoting it to a non-normative note?

2 things (and one extra):

1. Guide client implementers by using positive, actionable wording:
```kotlin
// This is very close to what a client would write
if (status == 2xx || content-type == "graphql-response+json") { 
  parseResponse() 
}

// I find this harder to read
if (status != 2xx && content-type == "application/json") { 
  // What to do here? I don't know, it depends.
} else {
  // This is actually the important branch
}
```
2. Keep the normative parts of the spec [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) by not repeating them (repetition in non-normative notes is ok). Took me a while to process that `MUST NOT` because I was trying to understand if this was new information or just a repetition.
3. Be more explicit about the rationale behind the 2 content-types. This is partially adressed in this PR by adding an intro sentence (`In certain environments,...`). Overall, I feel like `graphql-response+json` is a very important design choice. There are different notes in to the spec that address this but I found it hard to reconciliate all of them. By adding the intro sentence, I can link client implementers to that section and it's relatively self-contained.

Shane32: > What problem are we trying to solve by removing this or demoting it to a non-normative note?

I'll grant that the spec does contain a lot of client specifications, and written as such.

However, I feel the requirement in question simply ** is ** non-normative by nature, and probably not within scope of this document.

Let's compare to CORS issues.  Is CORS part of the GraphQL-over-http protocol?  No.  Is it part of the bigger picture that a client/server should consider when designing a server?  Yes.  How do we deal with it?  In #303 we add a non-normative note that mentions that if CORS issues are not considered, there may be security issues with the implementation.  Why isn't it normative?  CORS may not be relevant.  CORS may be solved in another fashion.  We don't require HTTPS either - for example, many microservices run on HTTP internally and the HTTPS layer is applied at the edge.  Perhaps we could add a non-normative note saying that data over HTTP is insecure, although the readers probably already know that and have a plan to keep their data secure.

By comparison, here we have an issue where an intermediary server might respond with failure.  Could this be relevant?  Yes.  Is it always relevant?  No.  Perhaps it is known that no intermediary servers are in use.  Perhaps the intermediary servers are configured to return 500 rather than 400 errors.  Perhaps they return xml instead of json responses.  Ultimately, I don't think this is part of the GraphQL-over-http protocol.  A note is warranted as a reminder to implementers.  

Now if it's deemed important enough to add a normative note, then it should be worded as an optional requirement with specific behavior, such as:
- A client SHOULD ignore any non-200 response that has an `application/json` content type, treating the response as a network error.

This tells the client exactly what to do, but as an optional requirement.  (just an example; it's worded poorly)

> When I read that as a client developer, "I MUST NOT trust something", I'm not sure what this concretely mean. If I get a 2xx or a graphql-response+json then I know what to do (parse the GraphQL response), if not then 🤷 I must not trust it but how does that concretely materialize? I'm not sure.

Completely agree.

> Guide client implementers by using positive, actionable wording

Agree

> Keep the normative parts of the spec [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) by not repeating them

Agree

> Be more explicit about the rationale behind the 2 content-types. This is partially adressed in this PR by adding an intro sentence (In certain environments,...). Overall, I feel like graphql-response+json is a very important design choice. There are different notes in to the spec that address this but I found it hard to reconciliate all of them. By adding the intro sentence, I can link client implementers to that section and it's relatively self-contained.

Yes, I think this PR does a better job communicating with the reader the logic/rationale of the altered content type and how to use it.

benjie: > I'm not sure what this concretely mean

I think "do not"'s are okay in specs - especially when you're telling someone to not do something that they typically would do unless told otherwise. But yes; we currently leave them to fend for themselves in this regard because there is no "correct" way to handle this, though there is an "incorrect" way, which is to blindly trust it.

> A client SHOULD ignore any non-200 response that has an application/json content type, treating the response as a network error.

I'm happy changing to wording along these lines - it aligns with what I said before: "throwing an error is the most reliable way of handling it; but we're not going to require that (though we could RECOMMEND it, I suppose)" - SHOULD and RECOMMEND have the same normative meaning.

I still think that we should pair these together though: you MUST NOT blindly trust it, and we RECOMMEND that you (i.e. "you SHOULD" - same difference) raise an exception as if it were a network error.

> Keep the normative parts of the spec [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) by not repeating them

I agree in general, but this is a conformance requirement specific to `application/json` so there should be a constraint in the `application/json` section - is that actually repeated?

martinbonnin: Reviving this thread. I've added back the `MUST NOT` compliance requirement [here](https://github.com/graphql/graphql-over-http/pull/304/commits/539dcd1599bcb7fcc75b66c851644af10ededcaa) as I think the non-normative "processing a response" section has some value and would like to see it included. We can discuss the value of the `MUST NOT` in follow up PRs.

@benjie let me know how that sounds.

benjie (approved):

PR #305, opened 2024-08-06
Persisted Documents: encourage URL approach

benjie:
This is just an early draft; I ran out of time.

### DIFF
--- a/spec/Appendix A -- Persisted Documents.md
+++ b/spec/Appendix A -- Persisted Documents.md
@@ -32,6 +32,51 @@ thus should generate different document identifiers.
 A _document identifier_ must either be a _prefixed document identifier_ or a
 _custom document identifier_.
 
+A _document identifier_ must only contain colons (`:`) and characters that are
+defined as
+[`unreserved` in RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3)
+(alphanumeric characters (`A-Z`, `a-z`, `0-9`), dashes (`-`), periods (`.`),
+underscores (`_`), and tildes (`~`)).
+
+DocumentIdentifier ::
+
+- PrefixedDocumentIdentifier
+- CustomDocumentIdentifier
+
+PrefixedDocumentIdentifier ::
+
+- UnreservedCharacter+ PrefixedDocumentIdentifierContinue+
+
+PrefixedDocumentIdentifierContinue ::
+
+- Colon UnreservedCharacter\*
+
+CustomDocumentIdentifier ::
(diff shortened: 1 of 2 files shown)


spec/Appendix A -- Persisted Documents.md
```diff
@@ -145,20 +225,62 @@ parameters in one of the manners described in this specification:
...
+such as caching, rate limiting, audit logging, access-pattern analysis, error
+detection, monitoring and more without needing to fully parse the incoming
+GraphQL request.
+
+For example, if the _GraphQL endpoint_ is `https://example.com/graphql` then a
+persisted document request may be made to an endpoint such as
+`https://example.com/graphql/sha256:517c56d2ba0779653b7698881207f749509f331bdaccbe951a82c378bc869556/FriendNames`.
+For documents containing a single anonymous operation the final segment must be
+omitted, e.g.
+`https://example.com/graphql/sha256:71f7dc5758652baac68e4a10c50be732b741c892ade2883a99358f52b555286b`.
```
n1ru4l, line 248: Should we maybe namespace persisted documents under `/graphq/<prefix>/<sha>`?

e.g.
- `/graphql/persisted/<sha>/<operation_name>`
- `/graphql/documents/<sha>/<operation_name>`
- `/graphql/persisted-documents/<sha>/<operation_name>`

Main reason for this is that `graphql-sse` already encourages `/graphql/stream`, which in theory can conflict with a "custom" hash. Changing this would also leave the options open to add other URL based features later on and not conflict with the persisted documents.

michaelstaib: We do namespace it in HotChocolate for less collision potential. However, Twitter for instance puts it directly on the GraphQL route as as outlined in this document.

n1ru4l: I'm not a big fan of having the operation name first. 🤔

From a server perspective, we first look up the hash for the document, and then the operation name within that document. Having the operation name first seems wrong to me.

From a debugging perspective, it does not matter too much whether the operation name comes before the hash.

n1ru4l: Also if we are looking at Chrome DevTools, the last URL part is displayed. So replacing operation name and hash might even be counter productive.

image

benjie: Good point, re Chrome devtools. Okay lets stick with the hierarchical approach, it makes more sense anyway :+1:

Harry5012 (approved):

PR #327, opened 2025-01-30
Require support of application/graphql-response+json by servers

Shane32:
As of 1/1/2025, the proposed watershed date has passed.  As the GraphQL over HTTP specification has not been released yet, eliminating the watershed period entirely may leave clients without adequate time to adjust to some of the new behaviors mandated by the specification.

This PR aims to separate the date that the server is required to support `application/graphql-response+json` from the date that it is required to be considered the default when no `Accept` header is provided.  This allows for clients to immediately take advantage of the benefits of the `application/graphql-response+json` media type for all compliant servers.

I'm suggesting that this support is required immediately.  While the original concept was that the spec would be released quickly, representing the current state of GraphQL servers, this did not occur.  However, the draft GraphQL-over-HTTP spec has been out since 2020, and has had no changes of substance (of any kind) since the addition of the `application/graphql-response+json` media type in Aug 2022.  As such, the fact that it was not "released as a spec" makes little difference.  Server implementations have had time to review and implement the new features of the draft specification, and various servers have done so.

Further, releasing the spec with this change does not break existing implementations.  They simply are not officially compliant until they support the new content type.  And finally, popular server implementations (assuming they have been watching this repository) have been expecting this change to take effect on 1/1/2025 and would expect to be non-compliant if they did not support the new content type.

So why make this change now?  I believe the sooner that all servers support the new media type, the sooner that the entire GraphQL ecosystem will adopt it.  Otherwise simple clients (e.g. handwritten with `fetch`) might not be able to use the new media type if servers are not supporting it, delaying general adoption.

This PR is complementary to pushing the watershed date further out as proposed within:
- #322

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -458,7 +458,7 @@ specific clients consuming their endpoint, thus both approaches are permitted.
 A server MUST support any _GraphQL-over-HTTP request_ which accepts the
 `application/json` media type (as indicated by the `Accept` header).
 
-A server SHOULD support any _GraphQL-over-HTTP request_ which accepts the
+A server MUST support any _GraphQL-over-HTTP request_ which accepts the
 `application/graphql-response+json` media type (as indicated by the `Accept`
 header).
 


enisdenjo (approved):
You've put it well in the description, I think this makes sense.

JoviDeCroock (approved):

Shane32:
> What are you seeing that simple clients will get out of this change?

I can only offer my own personal experience, which shapes my own viewpoint, and why I am advocating for the switch.

I started working with GraphQL around 6 years ago.  I was tasked with writing a high-performance website, and chose to use GraphQL over simple REST APIs as it met all of these objectives:
- Underlying database need only supply data for the requested fields, making use of indexes on large datasets
- HTTP transport need only supply data for the requested fields, lowering bandwidth requirements
- Complex queries (meaning, multiple inner joins) could be handled by the database directly, reducing latency between both the database and server, and more importantly, the server and client
- Very easy to develop with, and new fields/features do not (necessarily) break existing clients

However, the Apollo client for React had multiple major issues and could not be used.  First, it was very bulky, disproportionate to the rest of the application size.  Second, it was very slow, on two fronts: application startup incurred significant delays by Apollo processing the response (CPU time), and it injected __typename into every query, increasing bandwidth requirements.  And finally, it had problems with how it handled `resetStore`, breaking the login/logout portions of the site.

I have no doubt that by now Apollo or another library may work for me now, but at the time, I just needed and wrote some React wrappers for fetch, designing a lightweight GraphQL library with some basic caching features, which has worked well for me.

This brings me to the content type and response codes.  I wanted my server to be able to easily tell whether the result was a good request or not.  So I configured my servers to issue 4xx responses appropriately.  I like being able to review the HTTP logs and determine which requests are failing (typically auth errors) and not.  However, as many people have indicated, this can cause a conflict with HTTP relays and such returning 400 also.

As such, `application/graphql-response+json` seems like it would have been a better choice, had that been available.  If my testing of GraphiQL had noticed that `application/graphql-response+json` was used, I would have modeled my client library after it.  But at the time all the  software used `application/json`.  This is why I am pushing for the requirement now.  If the major GraphQL libraries had used the new content type, I would have modeled my client library after them.  If, on the other hand, everyone continues to use `application/json`, any developer writing their own implementation will just follow the pattern of everyone else.

As such, I recommend that `application/graphql-response+json` be required and `application/json` should in some way be discouraged while still being supported.  I just believe it's better for the community as a whole.

So, this PR simply requires support of the new content type.  It doesn't remove the requirement that `application/json` be supported.  In fact, with the current watershed date, it does not change anything at all in the spec.  But in general, favoring the new content type paves the way for new libraries to also do so.  The only situation the watershed breaks (which this PR itself does not change) are legacy clients that both (1) don't send an accept header, and (2) reject an `application/graphql-response+json` content type -- and that's assuming that the server is upgraded without the client being upgraded.

benjie:
What you describe explains exactly why this exists - so that you can see from the server logs what requests failed/succeeded - but it doesn't seem to answer the question as to how your _client_ itself would benefit?

> So, this PR simply requires support of the new content type. It doesn't remove the requirement that application/json be supported.

Indeed, but it _does_ allow clients to _stop_ supporting `application/json`, thereby dropping compatibility with existing (legacy) servers. That's what I think is a bad idea.

If we require all servers support the new content type, then taken at face value there's no reason for clients to support the legacy content type even though the code required to do so is minimal (i.e. to uncomment the commented part of this code: `if (res.headers.contentType === 'application/graphql-response+json' /* || (res.statusCode >= 200 && res.statusCode < 300) */) { handleTheResult(res.json) }` and to add `application/json` to their Accept header).

benjie:
Here's what I'm thinking: https://github.com/graphql/graphql-over-http/pull/330

benjie:
Accepted alternative solution https://github.com/graphql/graphql-over-http/pull/330 as discussed at the WG

PR #330, opened 2025-02-13
Remove watershed and prefer new format

benjie:
This is an alternative fix to #327 and #322 that does away with the watershed entirely. I'm coming to believe the watershed was a mistake.

We:

1. Require that compliant clients supply an `Accept` header (whilst noting that servers may support non-compliant clients).
1. Require that compliant clients include `application/graphql-response+json` in their `Accept` header
1. Name any server that only supports `application/json` a "legacy server"
1. Recommend that all new servers support `application/graphql-response+json`
1. Recommend that servers support `application/json` if and only if they want to support legacy (now non-compliant) clients

It is currently a draft because I've not had time to fully vet it, just wanted to get my thoughts down - I wouldn't review the specific wording too deeply yet.

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -465,8 +465,8 @@ responses.
 Each newly created or updated GraphQL _server_ SHOULD support responses using
 the `application/graphql-response+json` media type.
 
-:: A _legacy server_ is a _server_ that only supports responses using the
-`application/json` media type.
+:: A _legacy server_ is a _server_ that does not support responses using the
+`application/graphql-response+json` media type.
 
 Note: Prior to this specification, the media type `application/json` was in wide
 use for the HTTP response payload type. Unfortunately this means clients cannot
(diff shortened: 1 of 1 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -455,37 +455,34 @@ one of the media types it has requested, hence `406 Not Acceptable` being the
...
-`application/graphql-response+json` media type (as indicated by the `Accept`
-header).
+For maximal compatibility, a _server_ SHOULD support using both the
+`application/json` and the `application/graphql-response+json` media types for
+responses.
+
+Each newly created or updated GraphQL _server_ SHOULD support responses using
+the `application/graphql-response+json` media type.
+
+:: A _legacy server_ is a _server_ that only support responses using the
```
Shane32, line 468:
```suggestion
:: A _legacy server_ is a _server_ that only supports responses using the
```

Shane32 (approved):
This looks good to me.

enisdenjo (approved):
I like!

martinbonnin (approved):

martinbonnin:
Can we merge this one? I'd like to propose other changes but I'd rather not open too many PRs in parallel.

benjie:
I hacked it out quite quickly and didn't have time to proof-read it, I'm not sure it's currently consistent?

martinbonnin:
LGTM. I find that it improves the readability of the spec compared to the current state and I'd be in favor of merging it.

PR #332, opened 2025-02-21
Fix consistency of examples

AnthonyMDev:
Fix the Document Validation Failure examples to match the wording of the other examples.

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -600,7 +600,7 @@ code `200` (Okay).
 ##### Document validation failure
 
 If a request fails to pass _GraphQL validation_, the server SHOULD NOT execute
-the request and SHOULD return a status code of `400` (Bad Request).
+the request and SHOULD return a status code of `200` (Okay).
 
 ##### Operation cannot be determined
 
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -715,8 +715,8 @@ code `400` (Bad Request).
 
 ##### Document validation failure
 
-If a request fails to pass _GraphQL validation_, the server SHOULD NOT execute
-the request and SHOULD return a status code of `400` (Bad Request).
+If a request fails _GraphQL validation_, the server SHOULD return a status code
+of `400` (Bad Request) without proceeding to GraphQL execution.
 
 ##### Operation cannot be determined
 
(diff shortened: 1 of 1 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -599,8 +599,8 @@ code `200` (Okay).
 
 ##### Document validation failure
 
-Requests that fail to pass _GraphQL validation_, the server SHOULD NOT execute
-the request and SHOULD return a status code of `200` (Okay).
+If a request fails to pass _GraphQL validation_, the server SHOULD NOT execute
+the request and SHOULD return a status code of `400` (Bad Request).
```
Shane32, line 603: I think this should be 200 for application/json responses

AnthonyMDev: You're right. Thanks for catching that. Fixed now.

Shane32 (approved):
Fine with me

enisdenjo (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -715,8 +715,8 @@ code `400` (Bad Request).
 
 ##### Document validation failure
 
-Requests that fail to pass _GraphQL validation_ SHOULD be denied execution with
-a status code of `400` (Bad Request).
+If a request fails to pass _GraphQL validation_, the server SHOULD NOT execute
+the request and SHOULD return a status code of `400` (Bad Request).
```
benjie, line 719: This has turned from one compliance statement to two, both of which are "recommended". I don't think that's ideal - ideally we recommend that you "_both_ refuse to execute the request _and_ return a 400 status to say as much". Also now I come to think about it "fails to pass" is a bit redundant. Maybe something like:

```suggestion
If a request fails _GraphQL validation_, the server SHOULD return a status code of `400` (Bad Request) without proceeding to GraphQL execution.
```

benjie: NOTE: in traditional JSON we want two compliance statements here: should not execute, and SHOULD use 200, both of which are independent of each other (well... mostly. If you did execute then returning non-200 would be weird). You should always return 200, even on validation error; and you should not execute after validation error.

The same is not true with application/graphql-response+json - here the failure to execute and the 400 status are linked together. Failing to execute but returning non-4xx (e.g. 200) would be unexpected/undesired.

benjie (approved):

PR #370, opened 2026-03-29
Drop backwards compatibility

michaelstaib:


spec/GraphQLOverHTTP.md
```diff
@@ -183,10 +183,10 @@ using the JSON encoding for GraphQL requests:
...
 And for a _GraphQL response_:
 
-| Name                                | Description                                                        |
-| ----------------------------------- | ------------------------------------------------------------------ |
-| `application/graphql-response+json` | The preferred type for server responses; better HTTP compatibility |
-| `application/json`                  | An alternative type for responses (to support legacy clients)      |
+| Name                                | Description                              |
+| ----------------------------------- | ---------------------------------------- |
+| `application/graphql-response+json` | The preferred type for server responses   |
+| `application/json`                  | An alternative type for responses         |
```
martinbonnin: I would remove `application/json` altogether

spec/GraphQLOverHTTP.md
martinbonnin: Can we remove this?

martinbonnin (commented):
Quick cursory review. Couple of comments but looks good!

spec/GraphQLOverHTTP.md
```diff
@@ -249,28 +251,15 @@ A client MUST indicate the media types that it supports in responses using the
...
-`application/graphql-response+json`, it is RECOMMENDED that the client set the
-`Accept` header to `application/graphql-response+json`. Otherwise, to maximize
-compatibility the client SHOULD include the media type `application/json` in the
-`Accept` header and it is RECOMMENDED that the client set the `Accept` header to
-`application/graphql-response+json, application/json;q=0.9`.
-
-Note: The `q=0.9` parameter tells content negotiation that `application/json`
-should only be used if `application/graphql-response+json` is not supported.
+It is RECOMMENDED that the client set the `Accept` header to
+`application/graphql-response+json`.
```
martinbonnin: Do we  need this? There is a `MUST` just before and the explanation that it `MAY` add `application/json` just after so that sounds enough for me? 

```suggestion
```

BoD (approved):

benjie (changes requested):
For the reasons outlined in https://github.com/graphql/graphql-over-http/pull/379#issuecomment-4258953258 I do not think we should make this change.

PR #376, opened 2026-04-02
Add a section about the HTTP protocol

martinbonnin:
The current spec mentions HTTP 1.1 everywhere so I guess this is a requirement? But we might want to allow HTTP 2.0 still?

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -120,9 +120,7 @@ time.
 # HTTP protocol
 
 A server MUST comply with
-[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231) semantics. In this
-document, the words "header", "Content-Type" and "status code" refer to
-[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
+[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231) semantics.
 
 A server MUST support at least one version of the HTTP message format, such as
 [1.1](https://datatracker.ietf.org/doc/html/rfc7230) or
@@ -260,7 +258,8 @@ be part of a well-formed _GraphQL-over-HTTP request_.
 ## Accept
 
 A client MUST indicate the media types that it supports in responses using the
-`Accept` HTTP header.
+`Accept` HTTP header as specified in
+[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
 
 Note: If a client does not supply the `Accept` header then the server may
 respond with an error, or with any content type it chooses (including serving a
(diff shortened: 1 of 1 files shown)


benjie (changes requested):
I'm not convinced that we should require support for HTTP1.1. Is there a good reason that servers should not be able to be HTTP2+ only?

martinbonnin:
Note from meeting: for headers, mention both the 1.1 RFC and the 2.0 RFC, remove sentence from section 6.

Put it under "overview" so as not to change all the numbering.

spec/GraphQLOverHTTP.md
```diff
@@ -117,6 +117,25 @@ interact with a _GraphQL service_.
...
+Where applicable, server and clients MUST comply with
+[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231) semantics.
+
+In particular, "header", "status code", "body", "Accept", and "Content-Type" are
+to be interpreted according to
+[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
+
+A server MUST support at least one version of the HTTP message format, such as
+[HTTP/1.1](https://datatracker.ietf.org/doc/html/rfc7230) or
+[HTTP/2](https://datatracker.ietf.org/doc/html/rfc9113).
```
martinbonnin, line 134: Note from wg: remove this, do not bind ourselves to this.

spec/GraphQLOverHTTP.md
```diff
@@ -117,6 +117,25 @@ interact with a _GraphQL service_.
...
+
+A server MAY support several versions of the HTTP message format for wider
+compatibility.
```
martinbonnin, line 137: Note from wg: We can also remove that sentence, it's redundent with the above.

spec/GraphQLOverHTTP.md
```diff
@@ -117,6 +117,25 @@ interact with a _GraphQL service_.
...
 
+# HTTP protocol
+
+The details of the HTTP protocol used by GraphQL clients and servers are outside
+this specification.
+
+Where applicable, server and clients MUST comply with
+[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231) semantics.
+
+In particular, "header", "status code", "body", "Accept", and "Content-Type" are
```
martinbonnin, line 128: Note from wg: we should wrap those in spec-md formatting. 

cf https://github.com/graphql/graphql-over-http/blob/main/spec/metadata.json

benjie:
I _think_ this concern might be resolved by:

- #417

martinbonnin:
Works for me. Don't fix it if it ain't broken.

PR #398, opened 2026-05-05
Restore important guidance, improve legacy support

benjie:
In #379, support for `application/json` became optional, severely impacting the legacy of GraphQL clients and servers from the last decade.

This PR attempts to soften this blow by recommending that servers respond to `Accept: application/json` with `Content-Type: application/graphql-response+json` should they not wish to implement the `application/json` media type. For legacy clients that ignore the `Content-Type` of the response and only care about responses that contain `data`, this is a non-breaking change, allowing us to adopt and rely upon the new media type more widely.

The PR also performs editorial and restores some previous wording that I deem important - inline comments to follow.

One other key point is I try to recognize that there are/will be more media types the server will/can respond with (multipart, SSE, JSONL, etc) so some of the wording has changed to accommodate this.

### DIFF
--- a/spec/Appendix A -- application-json.md
+++ b/spec/Appendix A -- application-json.md
@@ -1,11 +1,13 @@
 # A. Appendix: `application/json` responses
 
+This section only applies when the response body uses or may use the legacy
+`application/json` media type for compatibility reasons.
+
 Previous to this specification, the article
 [Serving over HTTP](https://graphql.org/learn/serving-over-http)
 ([WayBack Machine entry, 1st June 2022](https://web.archive.org/web/20220601155421/https://graphql.org/learn/serving-over-http))
-on the graphql.org website served as guidance.
-
-This article used `application/json` as media type for the response.
+on the graphql.org website served as guidance. This article used
+`application/json` as media type for the response.
 
 In some cases, the response received by a client may not originate from a
 GraphQL service, but instead from an intermediary—such as an API gateway, proxy,
@@ -15,41 +17,43 @@ response with `4xx` or `5xx` status code and potentially using the standard
 `application/json` media type to encode the reason for the error. Such a
 response is unlikely to be a valid GraphQL response.
 
-For this reason, a client application receiving an `application/json` response,
-could rely on the response being a well-formed _GraphQL response_ only if the
-status code is `200`.
+For this reason, a client receiving an `application/json` response can only rely
+on the response being a well-formed _GraphQL response_ if the status code is
+`200`. However, responding to all GraphQL requests with HTTP 200 means that
+intermediary observability software cannot determine the status of the request
+without processing the response body.
 
-This caused multiple observability issues because it was challenging to
-distinguish a well-formed _GraphQL response_ from an intermediary response.
+Responding with the `application/graphql-response+json` media type enables the
+client to distinguish a well-formed _GraphQL response_ from an intermediary
+response, and is required by both clients and servers under this specification.
+This appendix exists to increase interoperability with legacy clients/servers
+that have not yet adopted this specification.
 
-`application/graphql-response+json` allows to distinguish a well-formed _GraphQL
-response_ from another intermediary response and is the required media type
-moving forward.
+Note: Servers may wish to respond to `Accept: application/json` requests with
+the `application/graphql-response+json` media type so that related HTTP tooling
+may utilize the HTTP status codes of responses without having to be
+GraphQL-aware. Doing so may impact error-handling behavior of legacy clients,
+and may prevent legacy clients from processing responses if they require the
+response `Content-Type` to be `application/json`.
 
-For compatibility reasons, clients and servers may support `application/json` as
-described in this section.
+## Accept
 
-Note: Servers may wish to only support the `application/graphql-response+json`
-media type so that related HTTP tooling may utilize the HTTP status codes of
-responses without having to be GraphQL-aware.
-
-### Accept
-
-To maximize compatibility, a client may include the media type
-`application/json` in the `Accept` header. When doing this, it is recommended
-that the client set the `Accept` header to
+To enable compatibility with legacy servers, the client SHOULD include the media
+type `application/json` in the `Accept` header and it is RECOMMENDED that such a
+client set the `Accept` header to
 `application/graphql-response+json, application/json;q=0.9`.
 
 Note: The `q=0.9` parameter tells content negotiation that `application/json`
 should only be used if `application/graphql-response+json` is not supported.
 
-### Status codes
+## Status codes
 
-When using the `application/json` media type, the server should use the `200`
-status code for every response to a well-formed _GraphQL-over-HTTP request_,
-independent of any _GraphQL request error_ or _GraphQL field error_ raised.
+When responding with the legacy `application/json` media type, the server SHOULD
+use the `200` status code for every response to a well-formed _GraphQL-over-HTTP
+request_ independent of any _GraphQL request error_ or _GraphQL field error_
+raised.
 
-If the response uses a non-`200` status code then the client must not rely on
+If the response uses a non-`2xx` status code then the client MUST NOT rely on
 the body to be a well-formed _GraphQL response_.
 
 Note: A status code in the `4xx` or `5xx` ranges or status code `203` (and maybe
@@ -60,14 +64,14 @@ _GraphQL response_ (because it cannot trust the source) the server must use
 generated or modified by an intermediary. See
 [processing a response](#sec-Processing-a-response) for more details.
 
-If the _GraphQL response_ contains a non-null {data} entry then the server must
+If the _GraphQL response_ contains a non-null {data} entry then the server MUST
 use the `200` status code.
 
 Note: This indicates that no _GraphQL request error_ was raised, though one or
 more _GraphQL field error_ may have been raised this is still a successful
 execution - see "partial response" in the GraphQL specification.
 
-The server should not use a `4xx` or `5xx` status code for a response to a
+The server SHOULD NOT use a `4xx` or `5xx` status code for a response to a
 well-formed _GraphQL-over-HTTP request_.
 
 Note: For compatibility with legacy servers, this specification allows the use
@@ -76,7 +80,7 @@ request_ where the response uses the `application/json` media type, but it is
 strongly discouraged. To use `4xx` and `5xx` status codes in these situations,
 please use the `application/graphql-response+json` media type.
 
-If the URL is not used for other purposes, the server should use a `4xx` status
+If the URL is not used for other purposes, the server SHOULD use a `4xx` status
 code to respond to a request that is not a well-formed _GraphQL-over-HTTP
 request_.
 
@@ -87,12 +91,12 @@ of `2xx` or `5xx` status codes when responding to invalid requests using the
 Note: URLs that enable GraphQL requests may enable other types of requests - see
 the [URL](#url) section.
 
-#### Examples
+### Examples
 
 The following examples provide guidance on how to deal with specific error cases
 when using the `application/json` media type to encode the response body:
 
-##### JSON parsing failure
+#### JSON parsing failure
 
 For example a POST request body of `NONSENSE` or `{"query":` (note: invalid
 JSON).
@@ -100,7 +104,7 @@ JSON).
 Requests that the server cannot interpret SHOULD result in status code `400`
 (Bad Request).
 
-##### Invalid parameters
+#### Invalid parameters
 
 For example a POST request body of `{"qeury": "{__typename}"}` (note: typo) or
 `{"query": "query Q ($i:Int!) { q(i: $i) }", "variables": [7]}` (note: invalid
@@ -109,25 +113,25 @@ shape for `variables`).
 A request that does not constitute a well-formed _GraphQL-over-HTTP request_
 SHOULD result in status code `400` (Bad Request).
 
-##### Document parsing failure
+#### Document parsing failure
 
 For example a POST request body of `{"query": "{"}`.
 
 Requests where the _GraphQL document_ cannot be parsed SHOULD result in status
 code `200` (Okay).
 
-##### Document validation failure
+#### Document validation failure
 
 If a request fails to pass _GraphQL validation_, the server SHOULD NOT execute
 the request and SHOULD return a status code of `200` (Okay).
 
-##### Operation cannot be determined
+#### Operation cannot be determined
 
 If [GetOperation()](<https://spec.graphql.org/draft/#GetOperation()>) raises a
 _GraphQL request error_, the server SHOULD NOT execute the request and SHOULD
 return a status code of `200` (Okay).
 
-##### Variable coercion failure
+#### Variable coercion failure
 
 If
 [CoerceVariableValues()](<https://spec.graphql.org/draft/#CoerceVariableValues()>)
@@ -146,7 +150,7 @@ For example the well-formed GraphQL-over-HTTP request:
 would fail variable coercion as the value for `id` would fail to satisfy the
 query document's expectation that `id` is non-null.
 
-##### Field errors encountered during execution
+#### Field errors encountered during execution
 
 If the operation is executed and no _GraphQL request error_ is raised then the
 server SHOULD respond with a status code of `200` (Okay). This is the case even
--- a/spec/Appendix A -- application-json.md
+++ b/spec/Appendix A -- application-json.md
@@ -65,7 +65,7 @@ generated or modified by an intermediary. See
 [processing a response](#sec-Processing-a-response) for more details.
 
 If the _GraphQL response_ contains a non-null {data} entry then the server MUST
-use the `200` status code.
+use a `2xx` status code and SHOULD use the `200` status code.
 
 Note: This indicates that no _GraphQL request error_ was raised, though one or
 more _GraphQL field error_ may have been raised this is still a successful
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -443,11 +443,13 @@ If the `Accept` header does not indicate support for one of the server's
 supported media types but does indicate support for `application/json`, the
 server MUST either:
 
-1. Respond with the `application/graphql-response+json` media type
-   (RECOMMENDED); OR
+1. Disregard the `Accept` header and respond with the
+   `application/graphql-response+json` media type (RECOMMENDED); OR
 1. Respond with the `application/json` media type as detailed in
    [Appendix A](#sec-Appendix-application-json-responses) (RECOMMENDED if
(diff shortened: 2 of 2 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -419,6 +416,13 @@ A server MUST indicate the media type of the response with a `Content-Type`
 header, and SHOULD indicate the encoding (e.g.
 `application/graphql-response+json; charset=utf-8`).
 
+A server MUST support responses using the `application/graphql-response+json`
+media type.
+
+:: A _legacy server_ is a _server_ that does not support responses using the
+`application/graphql-response+json` media type, and thus does not conform to
+this specification.
```
benjie, line 430: Moved this higher, introduced `legacy server` terminology.

spec/GraphQLOverHTTP.md
```diff
@@ -428,26 +432,41 @@ In alignment with the
...
+`application/graphql-response+json` but does indicate support for
+`application/json`, the server MUST either:
+
+1. Respond with the `application/graphql-response+json` media type
+   (RECOMMENDED); OR
+1. Respond with the `application/json` media type as detailed in
+   [Appendix A](#sec-Appendix-application-json-responses) (RECOMMENDED if
+   support for legacy clients is desired); OR
+1. Respond with a `406 Not Acceptable` status code and stop processing the
+   request (NOT RECOMMENDED).
```
benjie: This is the main change: we explicitly state that you should respond to `Accept: application/json` with `Content-Type: application/graphql-response+json` _or_ implement the appendix, and that responding 406 Not Acceptable is not recommended.

spec/GraphQLOverHTTP.md
```diff
@@ -428,26 +432,41 @@ In alignment with the
...
+trust responses from the server that do not use an HTTP `2xx` status code (since
+these replies may come from non-compliant HTTP servers or proxies somewhere in
+the request chain), hence the introduction of
+`application/graphql-response+json`. Legacy clients, utilities and tooling may
+continue to issue `Accept: application/json`, so to maintain interoperability it
+is recommended to continue to serve these requests.
+
+If the `Accept` header is present but does not indicate support for any of the
+server's supported media types, it is RECOMMENDED to respond with
+`406 Not Acceptable`.
```
benjie: This restores the previous `406 Not Acceptable` recommendation from the previous text, which we no longer apply to `application/json`.

spec/GraphQLOverHTTP.md
```diff
@@ -474,8 +493,31 @@ execution regardless of validation errors.
...
+(specifically {data} and {errors}) to determine the status of the response.
+
+Note: With `application/graphql-response+json`, clients know the response is
+well formed and should determine the detailed status of the response from the
+response body alone, allowing server authors to adopt more appropriate status
+codes without impacting behavior of existing clients. Intermediary servers may
+use the status code to determine the status of the _GraphQL response_ without
+needing to process the response body; this is useful in request logs, developer
+tooling, anomaly and intrusion detection, metrics and observability, API
+gateways, and more.
```
benjie, line 524: This is critical content that #379 dropped. If the client depends on status code then server admins cannot change status codes without risking breaking clients. For example, a naive server that responds to all non-data requests with `400` might not be able to later move to more nuanced codes such as `408 Request Timeout` or `429 Too Many Requests` if the client specifically checks status codes. Further, if we later expand this specification to allow responding with more mutation-centric status codes (`201 Created`, `202 Accepted`, `409 Conflict`, etc) then clients that depend on `response.statusCode === 200` may change behavior.

**It is vital for future expansion that the client should ignore status codes.**

spec/GraphQLOverHTTP.md
```diff
@@ -474,8 +495,31 @@ execution regardless of validation errors.
...
+Otherwise, the status code to use depends on the media type with which the
+GraphQL response will be served.
+
+For legacy `application/json` responses, see
+[Appendix A](#sec-Appendix-application-json-responses).
+
+### application/graphql-response+json
+
+This section only applies when the response body uses the
+`application/graphql-response+json` media type.
```
benjie, line 512: There's a strong likelihood of us supporting more media types over time (e.g. JSONL or multipart or SSE for incremental delivery), so I believe the section header is still warranted. Further; without it the text contradicts the appendix.

spec/GraphQLOverHTTP.md
martinbonnin: love the definition 👍. Should we also define `legacy client`?

spec/GraphQLOverHTTP.md
```diff
@@ -428,26 +434,41 @@ In alignment with the
 specification, when a client does not include at least one supported media type
 in the `Accept` HTTP header, the server MUST either:
 
+1. Disregard the `Accept` header and respond with the server's choice of media
+   type; OR
 1. Respond with a `406 Not Acceptable` status code and stop processing the
-   request (RECOMMENDED); OR
-2. Disregard the `Accept` header and respond with the server's choice of media
-   type (NOT RECOMMENDED).
```
martinbonnin: 👍

spec/GraphQLOverHTTP.md
martinbonnin: How do you recommend dealing with authorization? I have seen a fair amount of GraphQL APIs that rely on `401` to tell the client to refresh a token. Is that discouraged now?

spec/GraphQLOverHTTP.md
martinbonnin: Sounds good 👍

spec/GraphQLOverHTTP.md
martinbonnin: Sounds reasonnable to me.

martinbonnin (commented):
~~One~~ Two questions but looks good overall. Thanks for opening this!

spec/GraphQLOverHTTP.md
```diff
@@ -428,26 +434,41 @@ In alignment with the
...
+   request.
+
+If the `Accept` header does not indicate support for
+`application/graphql-response+json` but does indicate support for
+`application/json`, the server MUST either:
```
Shane32: This could be improved.  What if the request came in as `Accept: multipart/mixed, application/json;q=0.9`?  A server may wish to support `@stream` and `@defer` via `multipart/mixed` rather than responding along the lines below.

Shane32:
```suggestion
During response content negotiation, a server MUST support `application/json` as follows:
```

Shane32: Note that this paragraph and subsequent numbered items seems a little overstated, with little force.  Can we simplify it a bit?  The text above for the primary response type is clear and well stated (quoted below).

> A server MUST support responses using the `application/graphql-response+json`
media type.

The server _supports_ the type.  But the server is allowed to process the accept header as it sees fit considering any other formats it may support.  Maybe this paragraph and the subsequent list could be consolidated into a single sentence:

```suggestion
During response content negotiation, a server SHOULD support `application/json`
such that, when `application/json` is selected, the server responds either with
`application/graphql-response+json` or, when support for legacy clients is desired,
with `application/json` as detailed in
[Appendix A](https://github.com/graphql/graphql-over-http/pull/398/changes#sec-Appendix-application-json-responses).
```

(and removing the subsequent list)

There's no reason to say it MUST support it, if one of the options is to respond with 406.  And the server should be free to choose its own default format.

spec/GraphQLOverHTTP.md
Shane32: Note that this text here already indicates the _recommended_ behavior when the requested media type(s) are not supported.  So restating it in regards to `application/json` handling is redundant.

spec/GraphQLOverHTTP.md
benjie: If we refer to the term sufficiently to need it. Definitions like this only really matter when you reference them (via `_legacy server_` for example) as they become the target for when you click on the linked term.

spec/GraphQLOverHTTP.md
benjie: @Shane32 does it? I specifically removed the previous recommendations - these ones:

> In alignment with the [HTTP 1.1 Accept](https://tools.ietf.org/html/rfc7231#section-5.3.2) specification, when a client does not include at least one supported media type in the `Accept` HTTP header, the server MUST either:
>
> 1. Disregard the `Accept` header and respond with the server's choice of media type ~~(NOT RECOMMENDED)~~; OR
> 1. Respond with a `406 Not Acceptable` status code and stop processing the request ~~(RECOMMENDED)~~.

Did I mention it somewhere else?

benjie: OH! You mean because I told how to handle `application/json`... You're right, I missed the word "Otherwise" - will add.

benjie: I went a different way, the Otherwise didn't work well after a Note and moving the note didn't work well. Result:

> If the `Accept` header is present but indicates support for neither any of the server's supported media types nor `application/json`, it is RECOMMENDED to respond with `406 Not Acceptable`.

spec/GraphQLOverHTTP.md
benjie: A 401 with `application/graphql-response+json`? I've not seen that myself - normally that level of authentication happens at a higher middleware layer that doesn't yet process GraphQL (and is one of the contributing factors for the need for the `application/graphql-response+json` media type, since they commonly do respond in JSON for APIs).

Should the endpoint return a 401 with `application/graphql-response+json`, the `errors` array should indicate the authn issue too. I don't think there should ever be information in the status code that isn't present in the response body?

martinbonnin: I just checked the Buffer API and it's doing this:

```
$ curl -H "Accept: application/graphql-response+json" -v --json '{ "query": "{__typename}"}' https://api.buffer.com
< HTTP/2 401
< date: Wed, 06 May 2026 08:26:40 GMT
< content-type: application/graphql-response+json; charset=utf-8
<
{"errors":[{"message":"An authentication JWT or Access Token is required","extensions":{"code":"UNAUTHENTICATED"}}]}
```

Like you said, it also sets the error in the errors array. 

It's just painful as a client dev because every REST token refresh logic is using HTTP 401. I'm pretty sure lots of folks are relying on that right now as it's also easier than going into the errors array and checking extensions.

All in all, I wouldn't say clients should ignore the status code. They should not rely on it to determine whether the GraphQL response is well formed but it's still an information available to the client.

spec/GraphQLOverHTTP.md
benjie: You're right, this is overreaching and not what I intended; I've reworked the entire section.

benjie:
I woke up with a revelation: we can do a hybrid approach that increases compatibility whilst also maximizing use of status codes! For legacy requests (those that `Accept: application/json` and don't `Accept:` the new media type):

- successful and partially successful legacy requests can use `application/json` in the response because both media types require HTTP `2xx`
- unsuccessful legacy requests can use `application/graphql-response+json` with status codes, because they're already in a failure path anyway, so changing the "style" of error is not _as_ breaking as changing the happy path.

I've written this change up into this proposal, and if further reduces my concerns about breaking the legacy of GraphQL whilst being minimal effort for a server to support, not trading off the status codes that we're pushing for, and only being a non-normative recommendation.

spec/Appendix A -- application-json.md
```diff
@@ -60,14 +64,14 @@ _GraphQL response_ (because it cannot trust the source) the server must use
 generated or modified by an intermediary. See
 [processing a response](#sec-Processing-a-response) for more details.
 
-If the _GraphQL response_ contains a non-null {data} entry then the server must
-use the `200` status code.
+If the _GraphQL response_ contains a non-null {data} entry then the server MUST
+use a `2xx` status code and SHOULD use the `200` status code.
```
benjie, line 68: NOTE: this is a change from the previous spec, this is specifically to allow HTTP 294 responses (and other 2xx status codes) to legacy clients. AFAIK at least in the JS ecosystem most people use `response.ok` or check a status range rather than `response.status === 200` and this lines up better with the HTTP semantics - any 2xx is success.

Apollo client checks range: https://github.com/apollographql/apollo-client/blob/7c33689cca049f46192b091eaf979b2e64fd6213/src/link/http/parseAndCheckHttpResponse.ts#L151

URQL checks range: https://github.com/urql-graphql/urql/blob/1df98a66419b523bd8859ef617e13b4c7984bc7f/packages/core/src/internal/fetchSource.ts#L226
 
Graffle ignores status code: https://github.com/graffle-js/graffle/blob/4c58ee6cec67fe0047cc6708f6f19232077934e3/src/extensions/TransportHttp/TransportHttp.ts#L252-L260

Relay[^1] ignores status codes: https://relay.dev/docs/guides/network-layer/

[^1]: I guess I should say Relay's _network layer example recommended in the documentation_, since people keep helpfully pointing out Relay "has no network layer".

spec/GraphQLOverHTTP.md
```diff
@@ -239,11 +239,14 @@ client gets something useful, it needs to indicate the media types it supports.
 The client MUST include the media type `application/graphql-response+json` in
 the `Accept` header.
 
+:: A _legacy client_ is a client that does not support responses using the
+`application/graphql-response+json` media type, and thus does not conform to
+this specification.
```
benjie, line 244: Added definition for you @martinbonnin

spec/GraphQLOverHTTP.md
benjie: Good find - I wasn't aware of that usage!

The application can handle the status code _before_ passing it to the GraphQL client to be processed as a GraphQL response. The GraphQL client itself (URQL, Relay, Apollo Client, Graffle, etc) wouldn't normally be the one handling token refresh flows - that's normally a "link", "middleware", "exchange", or similar that the application developer inserts into the network chain and may not even process the response body.

Also: it's a SHOULD not a MUST. I think it's important that the GraphQL clients themselves ignore HTTP status codes, otherwise changing status codes could become a breaking change.

spec/GraphQLOverHTTP.md
benjie: I've reworked this whole section.

spec/GraphQLOverHTTP.md
martinbonnin: > - that's normally a "link", "middleware", "exchange",

Exactly. Apollo Kotlin does this using interceptors but it's the same concept really.

I guess there are 2 different mental models when it comes to clients:

1. Apollo Client is a "client"
2. A website is a "client"

In the context of this spec, I think of "client" mainly as 2.: everything that is one side of the network boundary, including links, interceptors, etc... 

Suggestion to remove the part about "status of the response". I don't think we have that language anywhere else and this lets type-2 clients deal with well-formed responses the way they want:


```suggestion
Clients should process the response as a well-formed _GraphQL response_
independent of the HTTP status code.

Note: With `application/graphql-response+json`, clients know the response is
well formed and should determine the detailed status of the response from the
response body alone, allowing server authors to adopt more appropriate status
codes without impacting behavior of existing clients. Intermediary servers may
use the status code to determine the status of the _GraphQL response_ without
needing to process the response body; this is useful in request logs, developer
tooling, anomaly and intrusion detection, metrics and observability, API
gateways, and more.
```

(That diff is large but only the first sentence is touched)

Shane32: Just FYI, GraphQL.NET Server responds with 401 and a well-formed GraphQL response (including an appropriate error) when the developer enables a schema-wide authorization policy but no authorization token is provided (or it is invalid), and 403 with a well-formed GraphQL response when the authorization policy is not met -- before parsing the document or validating the request.

Additionally, during the GraphQL document validation phase, if a referenced type/field requires authentication, or if a type/field requries an authorization role/policy not held by the client, the error is added to any other validation errors, and the status code is set to 401 / 403 as appropriate.

In this way, clients may examine the status code for 401/403 and handle it prior to the GraphQL client pipeline, as the execution has not yet begun.  Or they may process the response as a well-formed GraphQL response.

Shane32: I'd also like to remind us that we've designed it so 4xx within `application/graphql-response+json` means that the execution has not yet started.  So if the status code can provide any identifiable meaning to a client, it should be free to use that information and bypass processing the full GraphQL response.  Only 2xx means that there is data (full or partial) available.  Right?

Still fine to say that a GraphQL client should ignore the status code; if you're assuming that 'client' means 'client library' and excludes auth or other layers added by the developer.

Shane32:
I think I'm fine with this PR as-is, merging it back into #379 for final review.  Still seems a little over-stated but that's okay, and perhaps the extra text improves the reader's ability to understand the goals of the specification.

spec/GraphQLOverHTTP.md
martinbonnin: > if you're assuming that 'client' means 'client library' and excludes auth or other layers added by the developer.

If we do assume this, we should add a proper definition. My hunch is that it's going to be very hard to define.

benjie: You could say similar of server: is it everything the other side of the internet or is it just the GraphQL handling part. For server we’ve been explicit it’s just the GraphQL handling part and have explicitly called out that there are middlewares/intermediaries that may need to be traversed to get there (and that may behave differently). Maybe we need to either do the same for the client, or even broaden our definition so that middlewares/etc includes client-side middleware - they’re a pretty common fixture across clients so I think it’s reasonable to constrain the domain of client:

```mermaid
flowchart TB
  GraphQLClient --> ClientMiddleware
  ClientMiddleware --> Internet
  Internet --> ServerMiddleware
  ServerMiddleware --> GraphQLServer

```

```mermaid
flowchart TB
  GraphQLClient --> Middleware
  Middleware --> GraphQLServer
```

martinbonnin: > Maybe we need to either do the same for the client, or even broaden our definition so that middlewares/etc includes client-side middleware - they’re a pretty common fixture across clients

Works for me 👍 
I think this is worth being made explicit. That mermaid graph helped a lot. Would be cool to have it in the spec.

If we're being pedantic, there is also something at the top and bottom of this graph:

```mermaid
flowchart TB
  FrontendLogic --> GraphQLClient
  GraphQLClient --> ClientMiddleware
  ClientMiddleware --> Internet
  Internet --> ServerMiddleware
  ServerMiddleware --> GraphQLServer
  GraphQLServer --> BackendLogic
```

The `FrontendLogic` might use the HTTP status code. We have APIs in Apollo Kotlin to expose this. This was requested in multiple occasions.

I still think we should drop `and should read the response body (specifically {data} and {errors}) to determine the status of the response`. It introduces new language that is not really necessary IMO.

martinbonnin (approved):
LGTM, Thanks!

PR #404, opened 2026-06-23
Remove `application/json` response section; additional editorial

benjie:
- Fixes #379 
- Fixes #370

---

This pull request simplifies the GraphQL-over-HTTP spec by removing the sections on the legacy `Content-Type: application/json` _response_ type, and instead handling it through a recommendation: honour a request for `application/json` on successful requests, and use `application/graphql-response+json` for all other responses[^1]. This small change maintains broad (but not full) compatibility with legacy GraphQL clients without requiring an entire dedicated section (or appendix) detailing these rules.

Further, I've done a full review of the specification and made a few other adjustments based on expected upcoming needs.

[^1]: This is paraphrased, read the actual text for the specific wording.


spec/GraphQLOverHTTP.md
```diff
@@ -123,9 +115,6 @@ A _server_ MUST enable GraphQL requests to one or more GraphQL schemas.
 
 Each GraphQL schema a _server_ provides MUST be served via one or more URLs.
 
-A _server_ MUST NOT require the _client_ to use different URLs for different
-GraphQL query and mutation requests to the same GraphQL schema.
-
```
benjie, line 128: Removed: `/graphql/persisted/<hash>/<OperationName>` is intended in future. Didn't feel this rule was bringing much value

spec/GraphQLOverHTTP.md
```diff
@@ -327,8 +302,8 @@ http://example.com/graphql?query=query(%24id%3A%20ID!)%7Buser(id%3A%24id)%7Bname
 
 ## POST
 
-A GraphQL POST request instructs the server to perform a query or mutation
-operation. A GraphQL POST request MUST have a body which contains values of the
+A POST request instructs the GraphQL-over-HTTP server to perform a query or
+mutation operation. The request MUST have a body which contains values of the
```
benjie, line 306: Didn't like the ad-hoc definition of "GraphQL POST request" here.

spec/GraphQLOverHTTP.md
```diff
@@ -452,44 +434,33 @@ specification, when a client does not include at least one supported media type
...
-   type (NOT RECOMMENDED).
+   request; OR.
+1. Disregard the `Accept` header and respond with the server's choice of media
+   type.
+
+To improve compatibility with _legacy client_, if the `Accept` header does not
+indicate support for one of the server's preferred media types but does indicate
+support for `application/json`, it is RECOMMENDED to perform the request as if
+it had `Accept: application/graphql-response+json` but use
+`Content-Type: application/json` for any response with a `2xx` status code.
```
benjie: **This** is the key new paragraph. This normative paragraph, this recommendation, eliminates most of the need for the `application/json` section in the spec (whether that be in the spec or in an appendix).

From what I understand, this follows the behavior that @michaelstaib has already implemented in HotChocolate.

spec/GraphQLOverHTTP.md
```diff
@@ -689,11 +528,14 @@ reply with an appropriate `4xx` or `5xx` status code:
 
 - If the failure is due to an issue in the request itself, the appropriate `4xx`
   status code should be used:
+  - If a mutation is attempted via the `GET` verb, status code `405` MUST be
+    used.
```
benjie, line 535: Needed to mirror previous text, where this was MUST not RECOMMENDED.

martinbonnin: Any reason why this one in particular is a MUST when all the other status codes are RECOMMENDED?

martinbonnin (approved):
Looks good! 👍

PS: good job on filing this as PR #404 😄

benjie:
cc @Shane32 @PascalSenn @jerelmiller

spec/GraphQLOverHTTP.md
```diff
@@ -254,20 +236,13 @@ client gets something useful, it needs to indicate the media types it supports.
...
-should only be used if `application/graphql-response+json` is not supported.
+:: A _legacy client_ is a client that does not support responses using the
+`application/graphql-response+json` media type, and thus does not conform to
+this specification.
 
-The `application/graphql-response+json` media type adds improved support for
-HTTP status codes compared to the legacy `application/json` media type. When
-accepting both media types, the client SHOULD indicate it prefers
-`application/graphql-response+json` over `application/json`.
+If if is not known that the server supports `application/graphql-response+json`,
```
BoD:
```suggestion
If it is not known that the server supports `application/graphql-response+json`,
```

spec/GraphQLOverHTTP.md
benjie: Yes, because it is essentially a repeat of:

https://graphql.github.io/graphql-over-http/draft/#sel-EALJTCAACGzEr1S

> GET requests MUST NOT be used for executing mutation operations. If the values of query and operationName indicate that a mutation operation is to be executed, the server MUST respond with error status code 405 (Method Not Allowed) and halt execution. This restriction is necessary to conform with the long-established semantics of safe methods within HTTP.

I realised the bullet that follows this makes it seem like a SHOULD, so I added the explicit MUST to mirror that prior text.

I guess we could rewrite both places to use `SHOULD`, but it MUST be 4xx. I don't see much value in that though?

martinbonnin: It feels weird that we use MUST for 405 and not for the other status codes. I mean technically speaking, we could also use MUST for 408, 414. 

Doing it for 405 makes me thing there is a strong reason for using MUST there and not for the other ones. If there is no strong reason, I would favor consistency.

martinbonnin: > the server MUST respond with error status code 405 (Method Not Allowed) and halt execution

By the way, should the server return a valid GraphQL response in that case? (`content-type: application/graphql-response+json`)

spec/GraphQLOverHTTP.md
```diff
@@ -514,173 +485,41 @@ execution regardless of validation errors.
...
-
-Note: The purpose of setting a status code is to aid intermediary services and
-tooling (which may not implement this specification) in understanding the rough
-status of a response. This is useful in request logs, anomaly and intrusion
-detection, metrics and observability, API gateways, and more. The status code is
-not intended to aid the client, in fact it is recommended the client ignore the
-status code when this media type is in use.
+In case of errors that completely prevent the generation of a well-formed
+_GraphQL response_, the server SHOULD respond with the appropriate HTTP `4xx` or
+`5xx` status code depending on the concrete error condition.
```
phryneas: This could still be interpreted as "if the server cannot generate JSON at all", but doesn't say that in that case the `application/graphql-response+json` must not be used.

Do we need to define "well-formed"? It's used a lot, but not defined in this document.

spec/GraphQLOverHTTP.md
phryneas: I'd honestly say that this case is stronger than the others - in the other cases, there might be a hard requirement like compatibility with a weird proxy or something that steers towards a different response, but "mutation over GET" is such a strong no that it feels appropriate to me.
HTTP status codes are often kinda fluent, but sometimes they shouldn't be.

That said, I'd also support it if everything became MUST - but I'm not sure how many existing servers might violate it by now.

spec/GraphQLOverHTTP.md
```diff
@@ -689,11 +528,14 @@ reply with an appropriate `4xx` or `5xx` status code:
...
   - If the `Content-Type` of the request is not supported, status code `415` is
     RECOMMENDED.
-  - If none of the media types in the `Accept` header are supported, status code
-    `406` is RECOMMENDED.
+  - If none of the media types in the `Accept` header are supported and the
+    `Accept` header does not include `application/json`, then status code `406`
+    is RECOMMENDED.
   - If the client did not produce a request within the time that the server was
     prepared to wait, status code `408` is RECOMMENDED.
   - If the size of the URI was too large, status code `414` is RECOMMENDED (and
```
phryneas, line 544: Are we missing a `)` somewhere here? That `(and \`422\` is RECOMMENDED.` looks kinda out of place to me in general.

spec/GraphQLOverHTTP.md
martinbonnin: > "mutation over GET" is such a strong no that it feels appropriate to me.

FWIW, my suggestion here is to mandate 4xx or 5xx (as indicated in the paragraph just above). So "mutation over get" is still a strong no. 

405 is RECOMMENDED but if someone wants to use 499 or whatever, they technically can. But not 200, that is forbidden.

Reasonning is I like to explain this spec high level by saying "you can now use the status codes of your liking, you're not bound to 200 everywhere anymore".

If we have a MUST for 405, that undermines a bit that version because it's now "you can use the status codes of your liking BUT it MUST be 405 for this specific case". It weakens the general message IMO.

...unless there is a strong reason to disallow 4xx of course.

michaelstaib (approved):

spec/GraphQLOverHTTP.md
benjie: I don’t think “use the status codes of your liking” is a solid aim. Use the correct status codes for the situation; here’s the requirements, and for other situations there’s general recommendations to use but you can choose a better one if you know better.

spec/GraphQLOverHTTP.md
benjie: A well-formed GraphQL-over-HTTP request is one that meets the requirements of a “GraphQL-over-HTTP” request - i.e. https://graphql.github.io/graphql-over-http/draft/#sec-Request-Parameters . It’s not one purporting to do that (e.g. through media type) but then adding `variables: 7` which doesn’t meet the requirements, and thus is not well formed. If you don’t meet the basic requirements of the request, the server may respond how it likes.

spec/GraphQLOverHTTP.md
benjie: You might be seeing a rendering bug? Here’s what I see:

Image

spec/GraphQLOverHTTP.md
benjie: For a well-formed response it’s the same thing… so if something goes wrong in your JSON serializer or your GraphQL engine or whatever then maybe you’re not able to create a well formed response. You’re right though, we should explicitly forbid the media type in this position.

spec/GraphQLOverHTTP.md
martinbonnin: Sure! In all cases, I don't want to divert from the main point of this PR. Let's address that question separately.

martinbonnin (approved):

Shane32 (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -452,44 +434,33 @@ specification, when a client does not include at least one supported media type
...
 
 1. Respond with a `406 Not Acceptable` status code and stop processing the
-   request (RECOMMENDED); OR
-2. Disregard the `Accept` header and respond with the server's choice of media
-   type (NOT RECOMMENDED).
+   request; OR
+1. Disregard the `Accept` header and respond with the server's choice of media
+   type.
+
+To improve compatibility with _legacy client_, if the `Accept` header does not
```
jerelmiller:
```suggestion
To improve compatibility with _legacy clients_, if the `Accept` header does not
```

Should this be plural since "legacy client" isn't a proper noun?

spec/GraphQLOverHTTP.md
```diff
@@ -452,44 +434,33 @@ specification, when a client does not include at least one supported media type
...
+indicate support for one of the server's preferred media types but does indicate
+support for `application/json`, it is RECOMMENDED to perform the request as if
+it had `Accept: application/graphql-response+json` but use
+`Content-Type: application/json` for any response with a `2xx` status code.
```
jerelmiller: > use `Content-Type: application/json` for any response with a `2xx` status code.

I'm reading this with 2 different interpretations and am unsure which is correct so perhaps this might be an opportunity to clarify the wording a bit to disambiguate what this means. I read this as either:
- All responses get a `application/json` media type with a `2xx` status code instead of `application/graphql-response+json`/`4xx`
- A response that would normally get a `2xx` status code, gets the `application/json` media type instead. Continue to return `4xx` for GraphQL responses in cases where `application/grahpql-response+json` would normally do so (the "any response with a `2xx` status code" is what makes me think there might be other status codes the server could return with valid GraphQL responses instead of just the middleware scenerio listed below) 

Which is it?

I _think_ its the former given my historical context and the paragraph following this (non `2xx` can't be trusted to be a valid GraphQL response), but I want to make sure its clear. If thats correct, perhaps reword it slightly similar to the following?


```suggestion
`Content-Type: application/json` with a `2xx` status code for the response.
```

If neither of these interpretations are correct, I'd appreciate a better understanding 🙂

spec/GraphQLOverHTTP.md
```diff
@@ -714,76 +556,69 @@ reply with an appropriate `4xx` or `5xx` status code:
     `422` is RECOMMENDED.
   - If the variable values cannot be coerced to match the operation's variable
     definitions, status code `422` is RECOMMENDED.
-  - If the client is not permitted to issue the GraphQL request then the server
+  - If the client is not permitted to issue the GraphQL request, then the server
     SHOULD reply with `401`, `403` or similar appropriate status code.
-  - If the server is a short and stout ceramic vessel, status code `418` is
-    RECOMMENDED.
+  - If the server cannot process the request due to being a short and stout
```
jerelmiller, line 564: 🍵

jerelmiller (approved):
Love the changes! "does not conform to this specification" is so much simpler. Appreciate the updates!

spec/GraphQLOverHTTP.md
martinbonnin: I snapped on a little bit on this as well but I believe the goal is to link to the definition. (with the underscores)

Looks like putting the `s` outside the underscore works:

Screenshot 2026-06-30 at 15 52 15

```suggestion
To improve compatibility with _legacy client_s, if the `Accept` header does not
```

benjie: In spec-md we generally refer to everything as singular even when plural makes more grammatical sense. I don't think there's a `_keyword_s` in the GraphQL spec and I'd rather not introduce it here without that precedent.

I went with a different solution: [`1628de9` (this PR)](https://github.com/graphql/graphql-over-http/pull/404/changes/1628de95332f3e48893ece3f8a67880a90115d37)

spec/GraphQLOverHTTP.md
benjie: It was the latter; is this any clearer?

https://github.com/graphql/graphql-over-http/blob/f7459ef71ede7bb750897e52d7efd204ace7fb52/spec/GraphQLOverHTTP.md#L441-L447

spec/GraphQLOverHTTP.md
jerelmiller: I like it. Thanks!

spec/GraphQLOverHTTP.md
jerelmiller: Yep I like that better. The "except" removes the ambiguity for me. Thanks!

benjie:
Thanks for the feedback everyone; I've done some small tweaks to the wording to add clarity (and sidestep the plurality problem). If you're happy, we can get this merged and present the GraphQL-over-HTTP spec to the GraphQL WG tomorrow ready to join the main spec for spec cut in September!

Please slap another ✅ on this even if it seems redundant. cc @michaelstaib @martinbonnin @Shane32 @jerelmiller @phryneas @BoD @PascalSenn

jerelmiller (approved):
Appreciate it!

Shane32 (approved):

martinbonnin (approved):

BoD (approved):

phryneas (approved):

michaelstaib (approved):

benjie:
I think we've reached general consensus on this so I'm going to go ahead and merge it - thanks everyone!

Issue #205, opened 2022-07-19
Clarify the use of HTTP GET and POST request


TomokiMiyauci:
from [GraphQL over HTTP Spec](https://graphql.github.io/graphql-over-http/draft/#sec-Request)
> A server MUST accept POST requests, and MAY accept other HTTP methods, such as GET.

For semantics and performance (caching, etc.), it seems to me that it is recommended to use `GET` for the query operation.

If the options are not equivalent, it is necessary to clarify which is more appropriate.

spawnia:
One possible concern I could see with `GET` requests is the limited length of a URL. Some web server defaults:
- [Apache: 8190 bytes](https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestline)
- [NGINX: 8 KB](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers)
- [Node: 80 KB](https://github.com/nodejs/http-parser/blob/ec8b5ee63f0e51191ea43bb0c6eac7bfbff3141d/http_parser.h#L65)

When fragments, unstripped comments, lots of variables and multiple levels of nesting involved, payloads can get quite large. A quick look at our application logs showed a few queries above 1000 characters.

TomokiMiyauci:
@spawnia 

[RFC 7230](https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#request.target)
> It is recommended that all HTTP senders and recipients support, at a minimum, It is recommended that all HTTP senders and recipients support, at a minimum, request-line lengths of 8000 octets.

That is certainly true.

What do you think about caching of query operation?
Caching using the `POST` method seems to be possible, but I am not familiar with it.

Some implementations have complex mechanisms for handling cache on the client side, but I doubt it is really necessary.

It would be helpful to have a description of caching in the Spec.

reschke:
There we go.

- Yes, QUERY (RFC 10008) solves that. What **_is_** a problem though is that there doesn't seem to be a media type that would identify GraphQL is requests.
- GET is truly REQUIRED (and so is HEAD). Of course the server is free to send a response saying "hey, this is just GQL endpoint)

martinbonnin:
> GET is truly REQUIRED (and so is HEAD). Of course the server is free to send a response saying "hey, this is just GQL endpoint)

@reschke can you ellaborate on that? What's wrong with a server that returns `405 Method Not Allowed` for both `GET` and `HEAD`?

[...]

PR #411, opened 2026-07-03
Add `QUERY` verb

martinbonnin:
See https://www.rfc-editor.org/rfc/rfc10008.html and [ecosystem adoption](https://gist.github.com/desiderantes/2c7e657649cb92672d68e580fb69aa1d)

`QUERY` is an idempotent cacheable request with a body that maps quite well to GraphQL queries.

This editorial duplicates a bit of text but since this is a spec and not a novel, I'd rather have duplication than missing information. Maybe the JSON encoding part could be factored in but not sure how.


Shane32:
How about a bit more DRY?  For example:

> ## QUERY
>
> A [QUERY](https://www.rfc-editor.org/rfc/rfc10008.html) request instructs the GraphQL-over-HTTP server to perform a query operation.
>
> A server MAY accept QUERY requests. If a server accepts QUERY requests, it MUST handle the request and response according to all requirements that apply to POST requests, except as otherwise specified in this section.
>
> QUERY requests MUST NOT be used for executing mutation operations. If the values of {query} and {operationName} indicate that a mutation operation is to be executed, the server MUST respond with error status code `405` (Method Not Allowed) and halt execution. This restriction is necessary to conform with the long-established semantics of safe methods within HTTP.

Then any future changes to the POST requirements are inherited by the QUERY requirements.  Or are there more differences?

martinbonnin:
> Then any future changes to the POST requirements are inherited by the QUERY requirements. Or are there more differences?

There is one difference in the paragraph that `POST` is required while `QUERY` is optional:

```
A server MUST support POST requests encoded with the `application/json` media
type (as indicated by the `Content-Type` header) encoded with UTF-8.
```

But that's something you call out at the beginning of your version. 

I think the "condensed" part is easier to read and maintain but is a bit less clear. 

Both work for me.

dotansimha (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -403,6 +403,70 @@ And the body:
...
+## QUERY
+
+A [QUERY](https://www.rfc-editor.org/rfc/rfc10008.html) request instructs the
+GraphQL-over-HTTP server to perform a query operation. The request MUST have a
+body which contains values of the _GraphQL-over-HTTP request_ parameters encoded
+in one of the officially recognized GraphQL media types, or another media type
+supported by the server.
+
+A client MUST indicate the media type of a request body using the `Content-Type`
+header as specified in [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
```
benjie: Updated to 9110 RFC and referenced that this requirement actually comes from RFC10008.

```suggestion
As stated in
[IETF RFC 10008](https://datatracker.ietf.org/doc/html/rfc10008#name-query-method) a client MUST indicate the media type of a request body using the
`Content-Type` header; this header is specified in
[IETF RFC 9110](https://httpwg.org/specs/rfc9110.html#field.accept).
```

spec/GraphQLOverHTTP.md
```diff
@@ -403,6 +403,70 @@ And the body:
...
+A server MAY support QUERY requests encoded with and/or accepting other media
+types or encodings.
+
+If a client does not know the media types the server supports then it SHOULD
+encode the request body in JSON (i.e. with `Content-Type: application/json`).
+
+QUERY requests MUST NOT be used for executing mutation operations. If the values
+of {query} and {operationName} indicate that a mutation operation is to be
+executed, the server MUST respond with error status code `405` (Method Not
+Allowed) and halt execution. This restriction is necessary to conform with the
```
benjie: 405 would say that `/graphql` does not support the `QUERY` verb at all. Use 422 instead?

```suggestion
executed, the server MUST respond with error status code `422` (Unprocessable
Entity) and halt execution. This restriction is necessary to conform with the
```

martinbonnin:
Notes from July wg: 
* body size recommendations?
* mention [`Accept-Query`](https://www.rfc-editor.org/rfc/rfc10008.html#name-the-accept-query-header-fie)?

benjie:
Few notes:

- I think we're good with ignoring equivalent resource / HTTP 303?
- Normalization for caching may be an interesting consideration
- Should we recommend default cache directives?
- What does this mean for incremental delivery? I think it's fine: incremental delivery is cacheable.
- What does this mean for subscriptions? Potentially more challenging, since they're not really cacheable.
- 10008 states that QUERY is subject to CORS because `QUERY` isn't a safe-listed verb. But that's not a guarantee that it will never be a safe-listed verb... Do we need to consider adding some guidance such that if/when `QUERY` is safe-listed people don't get security issues?
- We should figure out some reasonable limits, or at least hint that you should have some. Might be a problem for the golden path, but definitely something implementers should think about.
- We should support `OPTIONS` so we can indicate `Accept-Query: application/graphql-request+json, application/graphql-variable-batched-request+json, application/graphql-persisted-request+json`
- Implementers `SHOULD` implement `QUERY`
- We should deprecate `GET`
- We should deprecate `query` over `POST`.

martinbonnin:
Thanks for the review! Comments inline

> I think we're good with ignoring equivalent resource / HTTP 303?

I think so too? But I might be missing some context.

> Normalization for caching may be an interesting consideration

Yup. Would recommend addressing outside this PR.

> Should we recommend default cache directives?

Sorry I missed that part of the discussion yesterday. The caching is outside this spec IMO? Browsers/CDN/etc.. can use HTTP caching with `QUERY` like they do for `GET` but this is HTTP behaviour, not GraphQL. I wouldn't necessarily address it here.

> What does this mean for incremental delivery? I think it's fine: incremental delivery is cacheable.
> What does this mean for subscriptions? Potentially more challenging, since they're not really cacheable.

We should definitely specify incremental delivery and subscriptions over websockets somewhere. I expect it to be a much larger endeavour than `QUERY` though.

> 10008 states that QUERY is subject to CORS because `QUERY` isn't a safe-listed verb. But that's not a guarantee that it will never be a safe-listed verb... Do we need to consider adding some guidance such that if/when `QUERY` is safe-listed people don't get security issues?

I'm not 100% clear on the impacts there. Do we have a documented security issue using `QUERY`? If not, I wouldn't necessarily warn users about something that is still hypothetical.

> We should figure out some reasonable limits, or at least hint that you should have some. Might be a problem for the golden path, but definitely something implementers should think about.

I'd prefer doing this in a separate PR. And probably address `POST` body length at the same time.

> We should support `OPTIONS` so we can indicate `Accept-Query: application/graphql-request+json, application/graphql-variable-batched-request+json, application/graphql-persisted-request+json`

100%

> Implementers `SHOULD` implement `QUERY`

Addresed in https://github.com/graphql/graphql-over-http/pull/411/changes/7a6701fd66117ca42b39a17f280b6d43b687d872

> We should deprecate `GET`

Agreed. Not 100% sure how to do it + `GET` is still useful for `application/graphql-request+json` so I haven't added language about it yet.

> We should deprecate `query` over `POST`.

Addressed in https://github.com/graphql/graphql-over-http/pull/411/changes/24d2675c48402ba2c3af0541be1dbbe52b8c7f96. I made this as a "Note", not sure how to deprecate otherwise.

spec/GraphQLOverHTTP.md
```diff
@@ -538,6 +610,8 @@ reply with an appropriate `4xx` or `5xx` status code:
   status code should be used:
   - If a mutation is attempted via the `GET` verb, status code `405` SHOULD be
     used.
+  - If a mutation is attempted via the `QUERY` verb, status code `405` SHOULD be
```
glen-84: Shouldn't this be `422` now (see above)?

If yes, `GET` should likely be changed as well?

martinbonnin: > Shouldn't this be 422 now (see above)?

Yup, 100%. This was a bad merge. Thanks for catching this 🙏 

> If yes, GET should likely be changed as well?

Good question. 405 has been around for a while and this hasn't been an issue. See also https://github.com/graphql/graphql-over-http/issues/205#issuecomment-5527325891. I'll defer to @benjie on this one.

PR #412, opened 2026-07-15
Suggestion to move HTTP wg to July 22nd

martinbonnin:
I won't be able to attend 30th of July and I think @benjie neither but there are things to discuss these days. Suggestion to move to 22nd of July.

I've added a couple of items to the agenda. @michaelstaib might be worth adding variable batching as well? 

Ping @benjie @Shane32 @michaelstaib


working-group/agendas/2026/07-Jul/23-graphql-over-http-wg-july-2026.md
benjie: You need to update the date/time.

benjie:
Rather than renaming the file, which will cause a new file to be created the next time the scripts run, create a new file for the new date/time and update the existing file to note it has been rescheduled.

22nd should work for me. I can't do 23rd or 30th.

martinbonnin:
Date adjusted in the calendar link and old file kept around.

PS: we'll also need to update the Google calendar.

benjie:
Yep, assuming this is merged before tomorrow that'll come up automatically in my Thursday scripts.

PR #416, opened 2026-08-03
RECOMMENDED to SHOULD

### LINKED ISSUE #409: SHOULD over RECOMMENDED
Noticed this in the latest spec text, ideally we use "SHOULD" instead of "RECOMMENDED" to simplify the prose alongside where "MUST" is used.

That would be more internally consistent since we use "SHOULD" elsewhere in the text, reducing the set of terms we use

(note: RECOMMENDED is a synonym for SHOULD https://datatracker.ietf.org/doc/html/rfc2119)

benjie:
- Fixes #409

Before merging this, merge:

- #417 

---

I am not particularly in favor of this change; but if we're going to do it we should do it right... I have put some normative changes in this, for example previously we RECOMMENDED that URLs end with `/graphql` but I'm not willing to say that they SHOULD end with `/graphql` (even though they have the same weight in RFC2119) so I've noted that it's a common convention instead.

I *would* recommend that people create their GraphQL APIs at `/graphql` or similar. Doing so _is_ recommended.

For someone who has deployed their GraphQL API at `https://example.com/api` or `https://graphql.example.com` I would **not** say that they "SHOULD NOT" have done that. It's fine.


spec/GraphQLOverHTTP.md
```diff
@@ -277,12 +277,12 @@ either omit {operationName} or set it to the empty string.
 GET requests MUST NOT be used for executing mutation operations. If the values
 of {query} and {operationName} indicate that a mutation operation is to be
 executed, the server MUST respond with an appropriate `4xx` status code and halt
-execution. Using `405` (Method Not Allowed) is RECOMMENDED. This restriction is
-necessary to conform with the long-established semantics of safe methods within
-HTTP.
+execution; status code `405` (Method Not Allowed) is RECOMMENDED. This
```
benjie, line 280: I've left this `RECOMMENDED` in place, it's the only one, but writing the sentence to use both MUST and SHOULD read really awkwardly and I believe RFC2119 gives us a choice of words to allow us to avoid these kinds of linguistic shenanigans.

spec/GraphQLOverHTTP.md
```diff
@@ -353,9 +353,8 @@ When encoded in JSON, a _GraphQL-over-HTTP request_ is encoded as a JSON object
   [the Response section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Response-Format.Response).
 
 All other property names are reserved for future expansion. If implementers need
-to add additional information to a request they MUST do so via other means; the
-RECOMMENDED approach is to add an implementer-scoped entry to the {extensions}
-object.
+to add additional information to a request they MUST do so via other means, for
+example by adding an implementer-scoped entry to the {extensions} object.
```
benjie: Adding to `extensions` is the "recommended" approach; but if you choose to set it via headers I wouldn't say you "should not" have done that. So I've lowered this to an example.

phryneas (approved):

martinbonnin (commented):
The `RECOMMENDED` wording grew on me. I've spent a couple of years with this spec using `RECOMMENDED` and it worked fine by me. All in all no strong opposition but also no strong drive to change this. I'll abstain.

BoD (approved):

spec/GraphQLOverHTTP.md
```diff
@@ -134,8 +134,8 @@ Server URLs which enable GraphQL requests MAY also be used for other purposes,
 as long as they don't conflict with the server's responsibility to handle
 GraphQL requests.
 
-It is RECOMMENDED to end the path component of the URL with `/graphql`, for
-example:
+It is a common convention to end the path component of the URL with `/graphql`,
+for example:
```
alf, line 138: A lot of graphql security hardening guides argue against using `/graphql`. My take is that if you're relying on the fact that your endpoint is non-standard for security then you're at a bad place. 

Still, I don't really see the benefit of recommending the path component. While it's all about discoverability for developers/tooling, it isn't something we can rely on. Saying that this is a common convention is better than recommending that implementers follow the convention. Good change.

spec/GraphQLOverHTTP.md
```diff
@@ -241,7 +241,7 @@ the `Accept` header.
 this specification.
 
 If it is not known that the server supports `application/graphql-response+json`,
-it is RECOMMENDED that the client set the `Accept` header to
+the client SHOULD set the `Accept` header to
 `application/graphql-response+json, application/json;q=0.9`.
```
alf, line 245: Why shouldn't the client set the `Accept` header to `application/graphql-response+json, application/json;q=0.9` if it is known that the server supports `application/graphql-response+json`? 😸

spec/GraphQLOverHTTP.md
alf: This read better with RECOMMENDED.

A possible direction might be to point out that this is one of the reasons the `extensions` object exists and link to the spec for more informations. That would be more informative to users new to the spec.

alf (approved):
I have no strong opinions, both before and after looks good to me.

spec/GraphQLOverHTTP.md
benjie: If it knows the server supports the desired media type, why `Accept` a worse one when you know it should never be used?

alf: My point is that the client shouldn’t really care. The result will be the same. The server will honor the preferred media type and the client can just use the recommended accept header.

spec/GraphQLOverHTTP.md
leebyron: yeah "common convention" is good, agree

spec/GraphQLOverHTTP.md
leebyron: reads correct, yes agree

michaelstaib (approved):

spec/GraphQLOverHTTP.md
leebyron: agreed -- a should/recommended (either) would clarify that extensions is the strongly preferred mechanism and not just an option

spec/GraphQLOverHTTP.md
```diff
@@ -354,9 +354,8 @@ When encoded in JSON, a _GraphQL-over-HTTP request_ is encoded as a JSON object
   [the Response section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Response-Format.Response).
 
 All other property names are reserved for future expansion. If implementers need
-to add additional information to a request they MUST do so via other means; the
-RECOMMENDED approach is to add an implementer-scoped entry to the {extensions}
-object.
+to add additional information to a request they MUST do so via other means, for
```
leebyron: this would read cleaner as what you "MUST NOT" do. you MUST NOT add additional top level properties?

leebyron (approved):
you can merge this as is if you like! feedback is for next action on the JSON props

spec/GraphQLOverHTTP.md
```diff
@@ -353,10 +353,10 @@ When encoded in JSON, a _GraphQL-over-HTTP request_ is encoded as a JSON object
...
   [the Response section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Response-Format.Response).
 
-All other property names are reserved for future expansion. If implementers need
-to add additional information to a request they MUST do so via other means; the
-RECOMMENDED approach is to add an implementer-scoped entry to the {extensions}
-object.
+All other property names are reserved for future expansion. Implementers MUST
+NOT add additional properties to the request object; additional information
+should instead use other means, for example HTTP headers or an
+implementer-scoped entry in the {extensions} object.
```
benjie, line 359: Discussed at WG, @leebyron felt we should use `MUST NOT` and then outline both approaches that could be used: extensions and headers. I have adopted this change.

spec/GraphQLOverHTTP.md
benjie: We came up with an alternative direction in the WG, hope that works for you!

PR #417, opened 2026-08-03
Consistency of reference to IETF RFCs

benjie:
Editorial and updated reference:

- RFC9110 replaces RFC7231 and covers more HTTP protocols (1.1, 2, 3)
- When referring to IETF RFCs, do so with a consistent format (`IETF RFC ####`)
- Hyperlink to the relevant sections where appropriate
- Clearly separate the GraphQL-over-HTTP spec's requirements ("A client MUST indicate the media type of a request body using the `Content-Type` header") from that of the IETF HTTP Spec (which uses SHOULD for this header).
- Fix the wording around the `Accept` header

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -646,7 +646,8 @@ response; it still indicates successful execution.
 # Non-normative notes
 
 This section of the specification is non-normative, even where the words and
-phrases specified in IETF RFC 2119 are used.
+phrases specified in [IETF RFC 2119](https://tools.ietf.org/html/rfc2119) are
+used.
 
 ## Partial success
 
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -225,8 +225,8 @@ be part of a well-formed _GraphQL-over-HTTP request_.
 ## Accept
 
 A client MUST indicate the media types that it supports in responses using the
-`Accept` HTTP header as specified in
-[RFC7231](https://datatracker.ietf.org/doc/html/rfc7231).
+`Accept` HTTP header; this header is specified in
+[IETF RFC 9110](https://httpwg.org/specs/rfc9110.html#field.accept).
 
 Note: If a client does not supply the `Accept` header then the server may
 respond with an error, or with any content type it chooses (including serving a
@@ -282,7 +282,7 @@ necessary to conform with the long-established semantics of safe methods within
 HTTP.
 
 Note: If status code `405` is used then the `Allow` header must be included as
-per [RFC9110](https://httpwg.org/specs/rfc9110.html#status.405).
+required by [IETF RFC 9110](https://httpwg.org/specs/rfc9110.html#status.405).
 
 ### Example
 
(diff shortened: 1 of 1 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -433,9 +434,9 @@ header and attempt to encode the response in the highest priority media type
 listed that is supported by the server.
 
 In alignment with the
-[HTTP 1.1 Accept](https://tools.ietf.org/html/rfc7231#section-5.3.2)
-specification, when a client does not include at least one supported media type
-in the `Accept` HTTP header, the server MUST either:
+[IETF RFC 9110 specification of `Accept`](https://httpwg.org/specs/rfc9110.html#field.accept),
+when the request includes an `Accept` header and no supported media type is
+acceptable according to that header, the server MUST either:
```
benjie, line 440: I reworded this because `application/json` is not "included" in `application/*` and yet it is "acceptable" by it.

spec/GraphQLOverHTTP.md
```diff
@@ -645,7 +646,7 @@ response; it still indicates successful execution.
 # Non-normative notes
 
 This section of the specification is non-normative, even where the words and
-phrases specified in RFC2119 are used.
+phrases specified in IETF RFC 2119 are used.
```
alf: This is the only textual reference to an RFC. It should probably be a link as well?

spec/GraphQLOverHTTP.md
phryneas: All I read in that linked paragraph regarding server behaviour is "to specify their preferences" - it doesn't really prescribe different ways the server reacts. Maybe another section?

martinbonnin: @phryneas did you have any section name in mind? The current structure keeps all the media type considerations in a single place so for server implementors, it's handy.

It does duplicate a bit of RFC 9110 but it's fine by me if it saves navigating to a separate RFC. I'd even consider re-using the language exactly:

```
As stated in [IETF RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-absence), 
if a content negotiation header field is present in a request and none of the available representations for the response can be considered acceptable according to it, the origin server can either honor the header field by sending a [406 (Not Acceptable)](https://datatracker.ietf.org/doc/html/rfc9110#status.406) response or disregard the header field by treating the response as if it is not subject to content negotiation for that request header field. This does not imply, however, that the client will be able to use the representation
```

martinbonnin (approved):

BoD (approved):

spec/GraphQLOverHTTP.md
benjie: @phryneas Maybe we should link to the content negotiation section instead?

https://httpwg.org/specs/rfc9110.html#content.negotiation

spec/GraphQLOverHTTP.md
benjie: I felt like this was already covered by the link at the top of the document; but yeah might as well

benjie: Fixed in [`9d61019` (this PR)](https://github.com/graphql/graphql-over-http/pull/417/commits/9d610198a22ccc75b35f2cb159a3288f63921c1c)

spec/GraphQLOverHTTP.md
benjie: Fixed in [`9bffebb` (this PR)](https://github.com/graphql/graphql-over-http/pull/417/commits/9bffebb52ef934ffc7027e5868721b8db0164177)

alf (approved):

PR #420, opened 2026-08-06
Fix "extensions" link

martinbonnin:
Looks like this link has moved

### DIFF
--- a/spec/GraphQLOverHTTP.md
+++ b/spec/GraphQLOverHTTP.md
@@ -198,7 +198,7 @@ parameters in one of the manners described in this specification:
   Operation.
 - {extensions} - (_Optional_, map): This entry is reserved for implementers to
   extend the protocol however they see fit, as specified in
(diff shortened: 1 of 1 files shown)


spec/GraphQLOverHTTP.md
```diff
@@ -198,7 +198,7 @@ parameters in one of the manners described in this specification:
   Operation.
 - {extensions} - (_Optional_, map): This entry is reserved for implementers to
   extend the protocol however they see fit, as specified in
-  [the Response section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Response-Format.Response).
+  [the Response section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Extensions).
```
benjie: Great catch; this looks like an unstable header though (any other header added for Extensions would cause it to be renamed) so let's stick with its parent:

```suggestion
  [the Response section of the GraphQL specification](https://spec.graphql.org/draft/#sec-Response-Format).
```

PR #423, opened 2026-08-07
Introduce GraphQL-over-HTTP response

martinbonnin:
Just like there is `GraphQL-over-HTTP request`, this PR introduces `GraphQL-over-HTTP response`. 

Using [GraphQL response](https://spec.graphql.org/draft/#sec-Response) was misleading because the HTTP specification doesn't say anything about response streams and we might end up using a separate media type for this (`application/graphql-responses+jsonl`, ...)

Instead, this PR introduces a new `GraphQL-over-HTTP response` that is either an execution result or a request error result, but not a response stream.


spec/GraphQLOverHTTP.md
```diff
@@ -493,94 +520,91 @@ execution regardless of validation errors.
...
-If the _GraphQL response_ contains the {data} entry and it is not {null}, then
-the server MUST reply with a `2xx` status code.
+If the _GraphQL-over-HTTP response_ contains the {data} entry and it is not
+{null}, then the server MUST reply with a `2xx` status code.
 
-If the _GraphQL response_ contains the {data} entry and does not contain the
-{errors} entry, then the server SHOULD reply with a `200` status code.
+If the _GraphQL-over-HTTP response_ contains the {data} entry and does not
+contain the {errors} entry, then the server SHOULD reply with a `200` status
+code.
```
magicmark: just curious why this is `SHOULD`? what other status code would be acceptable here? (in theory i could see something like a 201 on mutation, so i could see someone trying to be a good REST citizen and implementing that)

(I realize this is an existing wording and probably already discussed)

spec/GraphQLOverHTTP.md
```diff
@@ -406,17 +406,44 @@ And the body:
...
-any errors encountered during the request.
+a well‐formed _GraphQL-over-HTTP response_. The server's response describes the
+result of validating and executing the requested operation if successful, and
+describes any errors encountered during the request.
 
 A server must comply with
 [IETF RFC 9110](https://httpwg.org/specs/rfc9110.html).
 
+## GraphQL-over-HTTP response
+
```
magicmark: mentioned above:
> When encoded in JSON, a GraphQL-over-HTTP request is encoded as a JSON object

should this be updated to cover responses too

magicmark (approved):

spec/GraphQLOverHTTP.md
martinbonnin: 201 is the reason. Or more general make sure that clients to not do `== 200` checks in case we want to use other codes in the future.

Shane32 (approved):

Urigo (approved):

benjie:
@martinbonnin as agreed at the GraphQL-over-HTTP working group, I've updated this to lean heavily on the GraphQL spec to define the result shapes, rather than redefining them here. I've also fixed outdated references to terms in the spec that have since been refined - it wasn't just "GraphQL response" that was out of date (as you rightly caught!), but also "field error" has been replaced with "execution error" and the section links were also broken due to capitalization differences. I've moved to linking to the term definitions instead of section headers as I believe these links will be more stable. I've linked to the Sept 2025 main spec release right now, but if we can get a Sept 2026 release ready before this goes out I'll update the reference so they can go out either together or slightly staggered (HTTP following main spec) - that can be done in a follow up to this PR.

Please take a look and let me know what you think.

martinbonnin:
@benjie I can't approve my own PR but LGTM!

spec/metadata.json
```diff
@@ -1,12 +1,14 @@
 {
   "biblio": {
-    "https://spec.graphql.org/draft/": {
+    "https://spec.graphql.org/September2025/": {
```
martinbonnin, line 3: Was this meant to be commited? Do we have a release already? 👀

benjie: Wrong year, Martin!

> I've linked to the Sept 2025 main spec release right now, but if we can get a Sept 2026 release ready before this goes out I'll update the reference so they can go out either together or slightly staggered (HTTP following main spec) - that can be done in a follow up to this PR.

martinbonnin: 😂 What a year, huh?

Sounds good 👍 Thanks!

